PR #1, opened 2015-05-05
Test on travis

rleonid:


smondet:
The diff on the `META` file is very satisfying :)

LGTM !

PR #6, opened 2015-05-13
opam: dev-repo needs to end in .git

yomimono:
with a bare url, opam complains that this is not a recognised version-control URL.


rleonid:
Merged, Thank you!

PR #8, opened 2015-05-15
Add guards against instrumenting inside extensions

rleonid:
- modify the instrumenting logic to keep track of when it's copying an
  extension or attribute of the AST.
- add some tests cases against other ppx extensions (blob, getenv,
  deriving).
- remove commented out debugging code (added other).
- modify other tests expr_match.ml, source.ml that have no effect.


src/syntax/instrumentPpx.ml
```diff
@@ -354,26 +340,43 @@ class instrumenter = object (self)
           l
         in
           Str.value ~loc rec_flag l
-    | Pstr_eval (e, a) ->
+    | Pstr_eval (e, a) when not (attribute_guard || extension_guard) ->
```
smondet: Oh this one was well hidden! :+1:

smondet:
The diff LGTM, going to test on Ketrew's code  base now.

smondet:
It Works™ \o/

With branch https://github.com/hammerlab/ketrew/tree/add-bisect-ppx

and

```
opam pin add bisect_ppx https://github.com/rleonid/bisect_ppx.git#play_well_withothers
WITH_BISECT=true make distclean configure all
./ketrew-test ALL
make report
open _report_dir/index.html
```

src/syntax/instrumentPpx.ml
rleonid: That was the one causing all the fuss!

src/syntax/instrumentPpx.ml
```diff
@@ -354,26 +340,43 @@ class instrumenter = object (self)
...
-          match !InstrumentArgs.mode with
-          | InstrumentArgs.Safe   -> safe file
-          | InstrumentArgs.Fast
-          | InstrumentArgs.Faster -> faster file
-        in
-        header :: ast
-      else
-        ast
+    (*let ts = String.concat "," (List.map typoo ast) in *)
+    if extension_guard || attribute_guard then
```
aantron: @rleonid, I am currently reviewing some Bisect_ppx code. Does this test do anything in this function? What was the intent?

This is what I think happens: `extension_guard` or `attribute_guard` triggers `super#structure ast` on the next line (obviously). However, I believe it's redundant.

If `extension_guard` or `attribute_guard` is `true`, that means

1. We recursively traversed into the AST, so `get_filename ast` (or its alternative in the modern version of this code) did not evaluate to `None`.
2. The file has already been added to `InstrumentState`, as we are in a nested structure, and must have seen the file's top-level structure before.

That means that if we deleted this guard, the code in the `else` case would still fall through to `super#structure ast`, on line 380 at this point in history. The modern code does largely the same thing, just rearranged a bit.

Any insight?

Perhaps also @smondet on what the issue here was.

rleonid: I think this was necessary because there were two points that triggered instrumentation.

But you may also be right that it is no longer needed. I advise that you get rid of it and see if anything breaks, as we wrote tests for this. It would be faster to figure out if this line has meaning that way than me trying to recall 2 year old details.

aantron: Understood, thanks. It doesn't break any tests to remove it, so I'll remove it.

Issue #13, opened 2015-06-02
create separate findlib package for `bisect.cm[oax]`


smondet:
It would be nice to be able to `#require` (or compile aganist) instrumented libraries without dragging the `ppx` instrumentation.

PR #14, opened 2015-06-02
Toplevel support

rleonid:
@smondet now `#require "bisect_ppx.runtime";;` works fine in the toploop. Does this address #13 ?


smondet:
LGTM!

`bisect_ppx.runtime` works well for my use case.

`bisect_ppx` itself is still failing on the toplevel, it would be nice to still be robust is the package gets loaded by people:

```
# #require "bisect_ppx";;
/Users/sebastienmondet/.opam/4.02.1/lib/bisect_ppx: added to search path
/Users/sebastienmondet/.opam/4.02.1/lib/bisect_ppx/bisect.cma: loaded
/Users/sebastienmondet/.opam/4.02.1/lib/bisect_ppx/bisectThread.cmo: loaded
/Users/sebastienmondet/.opam/4.02.1/lib/bisect_ppx/./bisect_ppx.byte: activated
# let x = 42 ;;
Error: I/O error: //toplevel//: No such file or directory
```

rleonid:
I agree that would be a very useful warning. Any idea's on how to implement it?
It's not the actual loading of the runtime that's the problem, it's the activating of the `bisect_ppx.byte` to perform AST mapping.

smondet:
`bisect_ppx` could just ignore the toplevel silently?

rleonid:
Details? How would `bisect_ppx.byte` know which process (ocaml or ocamlc) is invoking it?

Or do you mean something different?

smondet:
I mean that when the file is `//toplevel//` just do nothing, copy the AST to the next `ppx`.

rleonid:
Ok, that seems reasonable.

rleonid:
With this change, do we still need a separate runtime lib, since the two packages would be pretty much the same?
`bisect_ppx.runtime` just wouldn't invoke the `ppx` mapper ?

smondet:
I think it's cleaner to have the `bisect_ppx.runtime` package (maybe you want to instrument your library, but not you tests and still compile them).

rleonid:
This fixes #13.

PR #17, opened 2015-06-11
Report correct bindings

### LINKED ISSUE #7: Fix Report tests
In `tests`
```sh
$ make one NAME=report
Running tests for 'report'...
     testing 'bisect' format ...
     testing 'csv' format ...
Files result.csv and reference.csv differ
     testing 'dtd' format ...
     testing 'dump' format ...
Files result.dump and reference.dump differ
     testing 'html' format ...
Files result.html and reference.html differ
     testing 'text' format ...
Files result.text and reference.text differ
     testing 'xml' format ...
Files result.xml and reference.xml differ
     testing 'xml-emma' format ...
```

The only difference in the `report` logic was replacing `String.create` with `Bytes.create`. But somewhere this broke. Really odd that the actual binary files match but not the custom formats.

rleonid:
@smondet Can you have a quick glance?
This fixes #7.


smondet:
The code LGTM.
I don't fully understand what's happening.

When you merge everything, before releasing, I can do more tests if you want (with Ketrew).

PR #20, opened 2015-06-16
Bugfix in CommentsLexer

hnrgrgr:
It currently can't lex:

```
let x = "\\\\" (* Miss the end of string *)
let y = '"'    (* Mistaken as start of string *)
let z = \"'    (* Mistaken as start of string *)
```

### DIFF
--- a/src/syntax/commentsLexer.mll
+++ b/src/syntax/commentsLexer.mll
@@ -61,6 +61,8 @@ let report_unmatched_pair () =
 let eol = ('\010' | '\013' |"\013\010" | "\010\013")
 
 rule normal ignored marked stack = parse
+| "'\"'"                    { normal ignored marked stack lexbuf }
+| "'\\\"'"                  { normal ignored marked stack lexbuf }
 | "\""                      { string 0 ignored marked stack lexbuf }
 | "(*BISECT-IGNORE-BEGIN*)" { let line = get_line lexbuf in
                               Stack.push line stack;
@@ -84,10 +86,8 @@ rule normal ignored marked stack = parse
 | _                         { normal ignored marked stack lexbuf }
 
 and string n ignored marked stack = parse
-| "\\\""                    { if n = 0 then
-                                normal ignored marked stack lexbuf
-                              else
-                                comment n ignored marked stack lexbuf }
+| "\\\\"                    { string n ignored marked stack lexbuf }
+| "\\\""                    { string n ignored marked stack lexbuf }
 | "\""                      { if n = 0 then
                                 normal ignored marked stack lexbuf
                               else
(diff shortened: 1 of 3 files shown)


rleonid:
Thanks for reporting, let me create a test case before incorporating your pull request.

rleonid:
> let z = \"'

Isn't valid `OCaml`. Did you mean `'\"'` ?

hnrgrgr:
On Tue, Jun 16, 2015 at 09:05:08AM -0700, Leonid Rozenberg wrote:
> > let z = \"'
> 
> Isn't valid `OCaml`. Did you mean `'\"'` ?

Indeed.

src/syntax/commentsLexer.mll
```diff
@@ -84,10 +86,8 @@ rule normal ignored marked stack = parse
 | _                         { normal ignored marked stack lexbuf }
 
 and string n ignored marked stack = parse
-| "\\\""                    { if n = 0 then
-                                normal ignored marked stack lexbuf
-                              else
-                                comment n ignored marked stack lexbuf }
+| "\\\\"                    { string n ignored marked stack lexbuf }
+| "\\\""                    { string n ignored marked stack lexbuf }
```
rleonid: I'm not certain why you need these? Can you provide a failing string?

src/syntax/commentsLexer.mll
```diff
@@ -102,6 +102,8 @@ and comment n ignored marked stack = parse
                                 normal ignored marked stack lexbuf
                               else
                                 comment (pred n) ignored marked stack lexbuf }
+| "'\"'"                    { comment n ignored marked stack lexbuf }
+| "'\\\"'"                  { comment n ignored marked stack lexbuf }
```
rleonid: Similar to the string case, if we're lexing a comment, the default pattern on line 110 seems to catch these.

rleonid:
I've added a branch [comment_lexer_bugs](https://github.com/rleonid/bisect_ppx/tree/comment_lexer_bugs) which has your reported errors. The first is currently handled correctly by the lexer, but I agree that you need the first 2 pattern matches in `normal` to deal with these special strings. 

Could you create some more examples that trigger the other parts of your fix (or remove them if they are not necessary)? 

Thank you!

hnrgrgr:
I've updated the PR with a new test.

The failling case was a string ```"a\"a"```. The current lexer terminates the string after ```"a\"```.

rleonid:
Now it makes sense to me, thanks for your help!

hnrgrgr:
Great news. This is a useful tool, thanks for holding the maintenance !

rleonid:
You're welcome. Though if you're interested in taking over, I'd be happy to pass the torch. My 'full'-time job is more working with data, rather than OCaml tool support.

PR #23, opened 2015-08-12
Fixing bugs in the fast mode implementation.

rleonid:
- Allocate a fixed size array at the beginning of the module,
  where counters are incremented during code execution.
- Updated the demo with fast targets.
- Update META to include a new package: bisect_ppx.fast.
- Changes to the test suite to accomodate changes.

This would fix #3

PR #24, opened 2015-08-13
Continue to resolve path when creating html reports.

rleonid:
This fixes a bug when files are implicit in the output file (bisect*.out) but actually in directories.


smondet:
OK, I tested a merge of this one with #23 and it works!

      bisect-ppx-report -I _build/src/pure/ -I _build/src/lib/ \
                        -I _build/src/app/ -I _build/src/test/  \
                        -I _build/ketrew-dummy-plugin-user \
                        -I _build/ketrew-integration-test \
                        -I _build/ketrew-test \
                        -I _build/ketrew-workflow-examples \
                        -verbose  -html _report_dir  bisect*.out

Issue #10, opened 2015-05-21
Results with ounit are incomplete


Khady:
test.ml
```ocaml
open OUnit2

let (===) = assert_equal ~printer: String.escaped

let test_foo _ =
  "foo" === "foo"

let test_bar _ =
  "bar" === "bar"

let test_baz _ =
  "baz" === "baz"

let test_qux _ =
  "qux" === "qux"

let test_suite =
  "example" >::: [
    "foo" >:: test_foo;
    "bar" >:: test_bar;
    "baz" >:: test_baz;
    "qux" >:: test_qux;
  ]


let suite =
  "all" >::: [
    test_suite;
  ]

let _ =
  run_test_tt_main suite
```

```
$ ocamlfind ocamlc -package bisect_ppx -package oUnit -linkpkg test.ml -o test.covered
$ ./test.covered 
....
Ran: 4 tests in: 0.10 seconds.
OK
$ bisect-report -html report_dir bisect0001.out
$ xdg-open report_dir/index.html
```

xml version:
```xml
<?xml version="1.0" encoding="iso-8859-1"?>
<!-- generated by Bisect 0.1 (http://bisect.x9c.fr) on 2015-05-21 19:25:29 -->
<bisect-report>
  <summary>
    <element kind="binding" count="4" total="7"/>
    <element kind="sequence" count="0" total="0"/>
    <element kind="for" count="0" total="0"/>
    <element kind="if/then" count="0" total="0"/>
    <element kind="try" count="0" total="0"/>
    <element kind="while" count="0" total="0"/>
    <element kind="match/function" count="0" total="0"/>
    <element kind="class expression" count="0" total="0"/>
    <element kind="class initializer" count="0" total="0"/>
    <element kind="class method" count="0" total="0"/>
    <element kind="class value" count="0" total="0"/>
    <element kind="toplevel expression" count="1" total="1"/>
    <element kind="lazy operator" count="0" total="0"/>
    <element kind="total" count="5" total="8"/>
  </summary>
  <file path="test.ml">
    <summary>
      <element kind="binding" count="4" total="7"/>
      <element kind="sequence" count="0" total="0"/>
      <element kind="for" count="0" total="0"/>
      <element kind="if/then" count="0" total="0"/>
      <element kind="try" count="0" total="0"/>
      <element kind="while" count="0" total="0"/>
      <element kind="match/function" count="0" total="0"/>
      <element kind="class expression" count="0" total="0"/>
[...]

PR #26, opened 2015-09-01
Open output file exclusively, no need for random.

rleonid:
To address #10 I had just created random integer suffixes to avoid the race condition. After putting on my thinking cap, this change reverses that style and files are open sequentially from different threads. 

cc @khady.


smondet:
LGTM !

PR #28, opened 2015-09-04
Parse OCaml line directives for BISECT comments

rleonid:
OCaml source code has line directives (# 1 "foo.ml") that change the
position, and more importantly for our case, filenames that are being
lexed. We would like to keep track of that information for correctly
implementing comments. The overall instrumentation though, is _still_
the top level filename because that strucutre is passed to the PPX
rewriter.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@ Version 0.2.5
 -------------
   - Revert the _Random_ indices for output files, just create multiple
   - Correct dependency on Str in META.
+  - Obey OCamls line directive when looking for BISECT comments.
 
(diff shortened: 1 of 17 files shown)


rleonid:
See http://caml.inria.fr/pub/docs/manual-ocaml/lex.html#linenum-directive for details.

smondet:
The diff looks good to me. But some documentation of the exact behavior of bisect_ppx Vs line-directives somewhere would be useful (e.g. in the README).

rleonid:
Thanks @smondet that's a good point.

PR #29, opened 2015-12-29
Add points on function expressions.

aantron:
In addition, I would suggest that the other tests in `expr_match.ml` be rewritten to avoid sequence expressions in cases. It seems that points are added on sequence expressions, and these mask potential problems with points being added to match cases.

### DIFF
--- a/src/syntax/instrumentPpx.ml
+++ b/src/syntax/instrumentPpx.ml
@@ -325,6 +325,9 @@ class instrumenter = object (self)
       | Pexp_match (e, l) ->
           List.map (wrap_case Common.Match) l
           |> Exp.match_ ~loc e
+      | Pexp_function l ->
+          List.map (wrap_case Common.Match) l
+          |> Exp.function_ ~loc
       | Pexp_try (e, l) ->
           List.map (wrap_case Common.Try) l
           |> Exp.try_ ~loc (wrap_expr Common.Sequence e)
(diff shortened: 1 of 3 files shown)


rleonid:
Thank you for the patch! I certainly haven't fully grokked the `OCaml` AST and I am worried about these missing corner cases.

Although, I did not write these original tests (they belong to Xavier Clerc) and can't speak authoritatively, I think that the sequence expressions make sense, and are testing for the wrapping each expression inside of the case list. Can you say more about the potential problem they might be masking?

rleonid:
Let me know if there are other issues or I can release a new version to opam. Again, thanks for the fix!

aantron:
Fair enough about the sequence expressions testing descent into cases. I suppose it's important to test that, so perhaps instead of rewriting the existing tests, it may be better to add new ones.

The potential problem that is being masked is this (came up in my testing):

```
function
| `A -> (* point *) e1; e1'
| `B -> (* point *) e2; e2'
```

As I understand it, a point is "inherently" generated "before" `e1` and `e2` because they are first in sequence expressions. If the point insertion code failed to add points "after" both of the `->` of the `function`, as before this commit, the test above would not detect that problem.

I don't believe such problems are currently present in the code. However, to catch regressions, it would be nice to have tests like this:

```
function
| `A -> e1
| `B -> e2
```

Where `e1` and `e2` are expressions that do *not* themselves result in points inserted. That would isolate point insertion for cases of `function`, and similarly for `match`. This is the reason I had to avoid sequence expressions inside the cases of the test that I added, for example.

As for corner cases, I think I might find a few more. I am currently applying bisect_ppx to a medium-sized project. You may want to wait a few days before pushing to OPAM as I work through writing tests for the code. I will let you know soon if I don't catch anything, otherwise I will send pull requests :)

PR #30, opened 2015-12-30
Add points on nested and anonymous functions.

aantron:
This commit adds points on nested and anonymous functions, but I have some reservations. I didn't write test cases because I want to discuss first, but my test program was:

```
let () =
  let f x = 0 in
  let g x = 1 in   (* I would like to know that this isn't called. *)

  f () |> string_of_int |> print_endline
```

Without this commit, points are not generated before `0` and `1`, and the coverage report does not indicate that `g` is not tested (or that `f` *is* tested, for that matter).

Concerns are:

- There is a separate function `wrap_func`, for wrapping `Pexp_fun`s that occur at the top level of a structure item binding or in a class. It seems the intent of the original code was only to generate points at the top level. Is there a reason for this?
- There isn't even a point kind for nested or anonymous functions (I used `Binding` for now). Again, it seems I am doing something in this commit that wasn't intended.

This change increased the number of points generated for my project by about 25% and slowed down testing just a bit.

### DIFF
--- a/src/syntax/instrumentPpx.ml
+++ b/src/syntax/instrumentPpx.ml
@@ -313,6 +313,9 @@ class instrumenter = object (self)
             List.map (fun vb ->
             {vb with pvb_expr = wrap_expr Common.Binding vb.pvb_expr}) l in
           Exp.let_ ~loc rec_flag l (wrap_expr Common.Binding e)
+      | Pexp_fun (al, eo, p, e) ->
+          let eo = map_opt (wrap_expr Common.Binding) eo in
+          Exp.fun_ ~loc al eo p (wrap_func Common.Binding e)
       | Pexp_apply (e1, [l2, e2; l3, e3]) ->
           (match e1.pexp_desc with
           | Pexp_ident ident
--- a/tests/ppx-instrument-fast/expr_binding.ml
+++ b/tests/ppx-instrument-fast/expr_binding.ml
@@ -15,3 +15,8 @@ let g x y z = (x + y) * z
 let g' x y =
   print_endline x;
   print_endline y
+
+let () =
+  let f _ = 0 in
+  let _g _ = 1 in
+  print_endline (string_of_int (f ()))
--- a/tests/ppx-instrument-fast/expr_binding.ml.reference
+++ b/tests/ppx-instrument-fast/expr_binding.ml.reference
@@ -1,5 +1,5 @@
 let ___bisect_mark___expr_binding =
-  let marks = Array.make 10 0
+  let marks = Array.make 15 0
   and (hook_before,hook_after) = Bisect.Runtime.get_hooks () in
   Bisect.Runtime.init_with_array "expr_binding.ml" marks true;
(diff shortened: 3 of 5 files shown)


rleonid:
You are correct, and this is a good catch. Can you add a simple test case (adding your test program to the end of `expr_match.ml` would be fine) and I'll merge?

Regarding your questions:
> There is a separate function `wrap_func`, for wrapping `Pexp_funs` that occur at the top level of a structure item binding or in a class. It seems the intent of the original code was only to generate points at the top level. Is there a reason for this?

I honestly don't know. I can make a hand-wavy argument about top level functions being more important. For example, they represent the values that you might export via an interface so you want to make sure that you've tested them. But I can see how it would also be useful to see local functions usage. I certainly use them all the time.

> There isn't even a point kind for nested or anonymous functions (I used `Binding` for now). Again, it seems I am doing something in this commit that wasn't intended.

I think that `Binding` is the right 'point kind' to use. I haven't used the point kinds for any analysis so I haven't thought too deeply about their usefulness or correctness. If you feel like you want to differentiate bindings at different levels, perhaps we can change the `Binding` variant to be `Binding of int` where `0` is top-level. That module has the framework to support versions.

aantron:
Regarding the point kind, I don't think it should be `Binding`. This change also affects expressions such as:

```ocaml
List.map (fun x -> (* point now added here *) x) l
```

Where there is no binding (of the function). Since point kinds are meant, in part, for toggling from the command line, it seems like the right thing to do is to add a new point kind, together with appropriate command-line option. Perhaps there are some projects that don't want a point on every anonymous function body. At the very least, "binding" is a confusing name for this behavior.

If you agree, I will amend this commit by:
- figuring out how to correctly distribute work between `wrap_func` and `method expr`,
- adding a new point kind and command-line option,
- adding tests,
- mentioning the option in the appropriate place in the manual.

rleonid:
You could argue that it is an 'anonymous' `Binding`. But the point now instrumented is in the evaluation of this binding, where there is a `Sequence` of expressions? Your current solution seems fine to me, so if you don't want to do all of those things, `Binding` or `Sequence` is fine with me. Otherwise, be my guest.

I don't think there is source for the manual. I've tried to keep track of the differences from `Bisect` in the [caveats section of the README.md](https://github.com/rleonid/bisect_ppx#caveats).

aantron:
I decided to go with your suggestion of reusing `Binding`, mainly because if I created a new kind such as `Function`, it would overlap with the current `Match`. I don't think `Binding` is the right name for this, but I decided to save sorting out the point kinds for some later time, if ever. Perhaps @xclerc can comment on point kinds and points in anonymous functions.

I added the tests for `expr_binding.ml` instead of `expr_match.ml`.

aantron:
It looks like this is the last pull request I will be submitting as a side effect of working on my current project, so if you'd like to release to OPAM, go ahead :)

rleonid:
Thanks again!

PR #31, opened 2015-12-31
Update bisect-ppx-report test for recent new points.

aantron:
My recent pull request #29 changed two binding points in `tests/report/source.ml` to match/function points, and I didn't notice that the `report` tests were failing as a result. This commit updates the reference files so that the tests succeed.

I also added a bunch of generated files to `.gitignore`.

The current pull request #30 doesn't affect the `report` tests (including in its amended version I have locally), so I think this one can and should be merged first.

For the future, it would be nice if test failure was more clearly indicated when running `make tests`.


aantron:
In particular, the test failure wasn't picked up by Travis either, since `make` exited with 0.

rleonid:
The one long standing remaining issue is to write a new test suite. The current one is showing its age.

PR #32, opened 2016-01-16
Opam dir

rleonid:
I think this should cure us of our `opam` flakiness.


rleonid:
The guide says that both things can be (inferred)[https://opam.ocaml.org/doc/Manual.html#opamfield-name].

aantron:
I think you should, because, as I understand it, inference is only possible when the `opam` file is in the repo. If someone (e.g. me) does `opam pin add .`, there are no repo directories to infer the name and version from.

PR #39, opened 2016-01-18
Untangle the build system

rleonid:
This commit starts to simplify the build system. To that end, I'd like to leverage ocamlbuild as much as possible. This commit:
  - Removed old ocamljava targets
  - Hide the output of ocamlbuild, it is still available in _build/_log
  - Perform file linking via ocamlbuild rules
  - Aggregate all of the build targets into one itarget

@aantron Let me know if there are other things that get in your way. My ultimate preference is a Makefile with simple targets, maybe a configure, and all the fun logic in `myocamlbuild.ml`.


aantron:
Looks basically good, just a few things:

- Spacing nit: https://github.com/rleonid/bisect_ppx/commit/efdeea4e515b71d6f8a3b7e68f202b7293581e90#diff-b67911656ef5d18c4ae36cb6741b7965R51
- Why remove ocamljava? I am not against removing it, but is it because we don't test on it anyway?
- :+1: for hiding ocamlbuild output, wanted to do the same thing.
- What is the new ocamlbuild rule doing? Can that not be done easier in the `Makefile`?
- We should probably add some kind of test(s) of the ocamlopt outputs. Right now I think we run only bytecode in tests. This will help to rewrite the build system more in the future and still be sure that we aren't making the ocamlopt output unusable in some way.

I'm going to hold off on #40 for now to avoid interference, since you are working on the build system. Feel free to do that issue too :) But if you don't, I'll take care of it eventually.

rleonid:
> Why remove ocamljava? I am not against removing it, but is it because we don't test on it anyway?
exactly.
> What is the new ocamlbuild rule doing? Can that not be done easier in the Makefile?
It creates the symbolic links from `report.byte` to `bisect-report` and `report.native` to `bisect-report.opt`. The `-no-links` gets rid of the standard link rules. It can be probably made easier with just the `Makefile` but I think that `myocamlbuild.ml` is a more flexible platform for writing build rules and I'd like to just keep them in one place.
> We should probably add some kind of test(s) of the ocamlopt outputs. Right now I think we run only bytecode in tests. This will help to rewrite the build system more in the future and still be sure that we aren't making the ocamlopt output unusable in some way.
Agreed.

Let me get #40 out of the way.

rleonid:
Current version fixes #40

aantron:
ocamlbuild is certainly more flexible, but it's also extremely verbose. While it would be nice, in principle, to have one build system for everything, but neither Make nor ocamlbuild are suited. I greatly prefer `ln` or `mv` in a `Makefile` to a bunch of lines in `myocamlbuild.ml` that just tell ocamlbuild to call `ln` anyway. `myocamlbuild.ml` is opaque. Especially as we clean up the `Makefile`, putting the `ln` in there will become much easier to deal with.

rleonid:
I agree that `ocamlbuild` is opaque, more so when we try to do complicated things. I think that your objects are valid. At the same time, I don't think that the things we're asking the build system to do (`ocamlbuild` or `make`) are necessarily that valuable. How about we just call the build targets `report.byte` and `report.native` and then change `opam` to install the correct thing?

aantron:
That sounds good to me. No need to symlink anything. I don't think the tests need to change because they already refer to the binaries and using their default names inside `_build`. May be overlooking something, however.

rleonid:
I want to do 2 more things:

1. Get the version logic to work again. Now that the `VERSION` file has been jettisoned, we'll use `git version` (via tag) or `opam` entry?
2. Look over the rest of those targets, such as the documentation generation.

aantron:
You mean something like `git describe --abrev=0`? That sounds good.

rleonid:
`make doc` works, and `ocamldoc` pages are generated in "ocamldoc". @aantron if this makes sense to you go ahead and merge.

aantron:
Looking great. Way cleaner and much more pleasant to run.

- I get this when building: `findlib: [WARNING] Interface topdirs.cmi occurs in several directories: /opt/local/lib/ocaml, /opt/local/lib/ocaml/compiler-libs`. Any ideas what is causing it? This happens on 4.02.2 and 4.02.3 switches (with different paths).
- Normally, I would be concerned about merging a PR with a large number of unsquashed commits, some of which are fixes of previous commits. But, what do you think about that? I can be fine with it :)
- Maybe we should choose to install the native rewriter instead of the bytecode one, it's supposed to run faster? It would go well with the `ocamlopt` testing we are about to start doing (so maybe we should wait until that's in `master`).
- Is it necessary to install `bisect_ppx.cmo`? I don't know enough about ppx rewriters, but it seems strange to me. I thought you need only the binary.

Some nits (if you want, I can fix them after the merge):

- `PROJECT_NAME` in the `Makefile` is not used anywhere now.
- There is a tab after `distclean` in the `Makefile` `default` target, that I just noticed because I'm using 2-space tabs in my editor. It looks fine in the terminal so it's not a big deal, but maybe we should add some kind of linter for the future to ban tabs. Would be a bit difficult in a `Makefile`, though.
- The doc generation has some warnings, but actually I fixed the underlying problem in a branch I haven't pushed. Please disregard them :)
- I'd delete the whole `ocamldoc` directory in `distclean`. It seems... cleaner.
- The `cd ..` at the end of `clean` is not necessary. The parens start a subshell. I'm also not sure why the `Makefile` tests for the sub-`Makefile`. We know it's there. I already simplified a similar mess in the `tests` target.

rleonid:
>I get this when building: findlib: [WARNING] Interface topdirs.cmi occurs in several directories: /opt/loca/lib/ocaml, /opt/local/lib/ocaml/compiler-libs. Any ideas what is causing it? This happens on 4.02.2 and 4.02.3 switches (with different paths).

That's a known `ocaml` [issue](http://caml.inria.fr/mantis/view.php?id=6754). I can't seem to get rid of it with warnings.

>  Normally, I would be concerned about merging a PR with a large number of unsquashed commits, some of which are fixes of previous commits. But, what do you think about that? I can be fine with it :)

Agreed, squashed.

> Maybe we should choose to install the native rewriter instead of the bytecode one, it's supposed to run faster? It would go well with the ocamlopt testing we are about to start doing (so maybe we should wait until that's in master).

Yea, I have no reason to pick one over the other, the bytecode rewriter's main advantage is being build slightly faster.

> Is it necessary to install bisect_ppx.cmo? I don't know enough about ppx rewriters, but it seems strange to me. I thought you need only the binary.

Yea. you're right. Looking through the original commits, I'm not certain why I added it. Removed.

> Some nits (if you want, I can fix them after the merge):

Addressed?

aantron:
:+1: LGTM. Any remaining nits I may find later, it's probably much more efficient for me to just fix instead of interrupting you :) It's a big cleanup.

PR #46, opened 2016-01-18
Add simple performance test

### LINKED ISSUE #45: Create a performance test
This would be not just for comparing modes against one another, but for understanding what effect modifications are having, and testing new modes proposed in #41.

aantron:
The test builds a small program instrumented in various modes and compares the resulting running times against each other and against an uninstrumented binary.

This currently uses the `time` command/shell built-in, so output may look different on each system. Might want to switch to something else later.

Resolves #45.

This was almost trivial to write with the new testing setup.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -19,27 +19,34 @@
 CFLAGS := -cflags -w,+A-4-48
 OCAMLBUILD_FLAGS := -use-ocamlfind -no-links $(CFLAGS)
 
-RUN := ocamlbuild $(OCAMLBUILD_FLAGS) test_main.native -- -runner sequential
+UNIT := ocamlbuild $(OCAMLBUILD_FLAGS) test_main.native -- -runner sequential
+PERFORMANCE := \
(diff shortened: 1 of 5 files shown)


aantron:
Here is sample output on my system.

```
 uninstrumented

real	0m0.018s
user	0m0.010s
sys	0m0.004s

 safe

real	0m1.105s
user	0m1.096s
sys	0m0.005s

 safe-threads

real	0m1.138s
user	0m1.128s
sys	0m0.005s

 fast

real	0m0.147s
user	0m0.140s
sys	0m0.004s

 fast-threads

real	0m1.333s
user	0m1.323s
sys	0m0.005s

 faster

real	0m0.114s
user	0m0.106s
sys	0m0.003s

 faster-threads

real	0m0.108s
user	0m0.100s
sys	0m0.004s
```

It seems that fast isn't really fast anymore when you need it to be thread-safe. I found that surprising, will check in more detail later for whether it's true and why.

rleonid:
Interesting.

rleonid:
See my comment on your pull request. Faster was Fast because it was always locking the mutex.

Issue #41, opened 2016-01-18
Reorganize modes -safe, -fast, -faster


aantron:
There are several issues here.

1. `-fast` is also apparently safe. @rleonid if `-fast` is faster than `-safe` (is it?), why is there `-safe`? Is `-safe` more correct than `-fast` in some case?
1. I want to write a little performance test so we can understand the real performance of the various modes.
1. I want to also write a race condition test.
1. The difference between `-fast` and `-faster` is that `-faster` doesn't try to take locks even if you link in `BisectThread`.
  - The reason for locks in either of these modes is that they increment counts. Are counts actually often used from Bisect_ppx? If we were only updating `true`/`false` flags, we could write `true` without locks in all cases. Maybe what we need instead of `-fast` and `-faster` is `-boolean` and `-numeric`, and if you choose `-numeric`, it is always correct, but you pay a price for locks.
  - If we don't want to do this, I want to look at the actual performance difference between `-fast` with locks and `-faster`. It's probably significant, but if it isn't, I would like to eliminate `-faster`.
  - There may also be a way to use the `ocaml.ppx.context` attribute, or some other method, to find out if we need to take locks. This would eliminate the brittleness of `BisectThread`, but also take per-module control over safety away from the user compiling the code. Does anyone actually take advantage of this control?

I'm currently hoping to make work `-boolean`/`-numeric` with automatic detection of the need for locks in `-numeric`. I'd probably want to make `-boolean` the default. It wouldn't affect the report tool. The runtime would just output counts 0 and 1.

rleonid:
Without quoting...
1. As far as I've tested it, yes, `-fast` is much faster than `-safe`. Around an order of magnitude (minutes -> seconds) on [Oml](https://github.com/hammerlab/oml) and [Ketrew](https://github.com/hammerlab/ketrew/). I think the `-safe` is mostly legacy, I've kept it around for backward compatibility. I've never found an instance where they diverged, but I've only spot checked the test suite and the projects where I've applied the switch. If you want to just have one mode that is canonical and "just works", I am all in favor of jettisoning `-safe` and changing `-faster.`
2. A performance test is a good idea, but not required. The improvement is noticeable.
3. A race condition test is probably a good idea, we had a bug dealing with `OUnit` threads a couple of months ago.
[...]

PR #47, opened 2016-01-19
Add thread safety test

aantron:
Part of #41.

The tester itself "should" be done, but I'm having trouble producing a program in which one thread actually pre-empts another during counting. @rleonid can you take a look at the `source.ml`?

Both threads are indeed running. The tester expects the count to be wrong here (not 2 million), but it comes to 2 million anyway. I don't know if it's the `Threads` implementation on my system, bad code on my part, or it's just that rare for pre-emption to happen there in that loop.

I inspected the output of `bisect_ppx` with `-dsource`, and it indeed doesn't take locks.

### DIFF
--- a/tests/thread-safety/source.ml
+++ b/tests/thread-safety/source.ml
@@ -1,0 +1,10 @@
+let () =
+  let repeat n = for i = 1 to n do () done in
+
+  let count = Sys.argv.(1) |> int_of_string in
+
+  let thread_1 = Thread.create repeat count in
+  let thread_2 = Thread.create repeat count in
+
+  Thread.join thread_1;
+  Thread.join thread_2
--- a/tests/test_main.ml
+++ b/tests/test_main.ml
@@ -29,6 +29,7 @@ let tests = "bisect_ppx" >::: [
   Test_exclude.tests;
   Test_exclude_file.tests;
   Test_ppx_integration.tests;
+  Test_thread_safety.tests;
   Test_ounit_integration.tests
 ]
 
--- a/tests/thread-safety/reference
+++ b/tests/thread-safety/reference
@@ -1,0 +1,1 @@
+    <point offset="44" count="2000000" kind="for"/>
--- a/tests/thread-safety/test_thread_safety.ml
+++ b/tests/thread-safety/test_thread_safety.ml
@@ -1,0 +1,60 @@
+(*
+ * This file is part of Bisect_ppx.
+ * Copyright (C) 2016 Anton Bachin.
+ *
+ * Bisect is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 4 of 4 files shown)


tests/thread-safety/source.ml
```diff
@@ -0,0 +1,13 @@
+let () =
+  let rec repeat n =
+    if n <= 0 then ()
+    else repeat (n - 1)
+  in
```
smondet: AFAIR OCaml threads can be preempted only if there is allocation.

And from experience, to make sure threads preempt each other doing some I/O would be best.

(maybe the code added by `bisect_ppx` is enough actually)

aantron:
> AFAIR OCaml threads can be preempted only if there is allocation.
>
> And from experience, to make sure threads preempt each other doing some I/O would be best.
>
> (maybe the code added by bisect_ppx is enough actually)

This is what I was starting to suspect. Actually, this means that Bisect_ppx is already safe with respect to `Threads` even without locks, because there is no allocation when incrementing an integer (obviously).

I can't think of any other way of getting fine-grained pre-emption in OCaml. What will multicore do with respect to `Threads`? Will they actually be running concurrently?

@rleonid The problem with OUnit you referred to was a problem with multiprocessing and not multithreading, right?

rleonid:
Yea, you're right the problem was multiprocessing. It took me a while to figure out what `OUnit` was doing under the hood.

tests/thread-safety/source.ml
rleonid: Why not use `Thread.yield` ?
```OCaml
else if n mod 100 = 0 then begin Thread.yield (); repeat (n - 1) end
```

aantron: The threads are interleaving, just not in the middle of updating a count. I would have to insert the `Thread.yield` call into the instrumentation, not into the code being instrumented.

aantron:
I suppose we could add a `-debug-yield` option to the instrumenter so insert `Thread.yield` calls into the instrumentation, but I want to hold off on that until I deal with the modes, because the modes PR will decrease the number of places that have to be modified to do this insertion.

I have an updated version of this PR that skips tests expected to fail (which don't, due to lack of pre-emption). It is rebased, and therefore waiting, on the `ocamlopt` PR, however.

rleonid:
Sounds good.

tests/thread-safety/source.ml
rleonid: Now I understand. Tough problem to test for.

aantron:
I think this version should go in. In light of https://github.com/rleonid/bisect_ppx/issues/41#issuecomment-172926582, I don't think there is any real use in inserting `yield` calls. We can wait until we either find a way to make `-faster` unsafe, or someone reports a bug. The thread safety test should then fail if we can reproduce the setup. We can then also uncomment https://github.com/rleonid/bisect_ppx/commit/65be899c034e89c9b98f42af28557eab9e3448a4#diff-bc26551674bb733600a6876b7240ffb9R33 to get a failure test.

rleonid:
:+1:

PR #49, opened 2016-01-19
More build system and metadata cleanup

aantron:
- Cleaned up `.gitignore`.
- Cleaned up `META` and added version number to `fast` subpackage. It looked weird without it in `ocamlfind list`.
- Put a compiler constraint in `opam`, so that in the extremely unlikely event we drop ppx_tools, or ppx_tools supports 4.01.0, Bisect_ppx does not become installable (yet broken).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,26 +1,13 @@
-*.annot
-*.cmo
-*.cma
-*.cmi
-*.cmp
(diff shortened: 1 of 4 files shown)


rleonid:
LGTM.

PR #50, opened 2016-01-19
One mode

aantron:
See the commit message in the first commit.

Note that whatever race condition may have existed here: https://github.com/rleonid/bisect_ppx/commit/dbe8a84879a80c07b899d76666522262b359fde8#diff-d347ee27660b0e186d9f79a1f8950eb1R44 might not exist anymore, because that code is now only called by `at_exit`, instead of on every unsafe mark.


rleonid:
Do you mind if I take a day or 2 to look this over? I need to allocate a solid amount of time for this that I don't have at the moment.

aantron:
Not at all, take your time.

rleonid:
I just realized I hadn't actually correctly implemented `Faster` in `ppx` at all.

aantron:
What do you mean? The code seemed true to its description, which is being like `-fast` but not thread-safe by not even trying to take locks.

rleonid:
This [line](https://github.com/rleonid/bisect_ppx/blob/master/src/syntax/instrumentPpx.ml#L196) should have `falseconst` for `Faster`.

aantron:
Yeah but IIRC that line only controls whether you get a verbose warning about being unsafe. Locking code wasn't generated in `-faster`, [this line](https://github.com/rleonid/bisect_ppx/blob/master/src/syntax/instrumentPpx.ml#L226) controls that.

aantron:
Oh, well we are also talking about different aspects of it. I guess the locking code was left in the startup/shutdown procedures with `-faster`, but it wasn't emitted for the counts, the latter of which is correct and more consequential IMO. You can see the output of `-faster` in the reference files in the first commit in this PR.

rleonid:
You're right again, still loading this code into context.

PR #51, opened 2016-01-20
Use Bisect_ppx on Bisect_ppx

aantron:
See the commit message.

This generates reports like

```
Finished, 41 targets (0 cached) in 00:00:02.
......................................................SSSSSS..SS....
Ran: 68 tests in: 14.41 seconds.
OK: Cases: 68 Skip: 8
 Coverage:
 - total: 1197/1598 (74.91%)
```

screen shot 2016-01-19 at 19 51 57

We can improve the HTML and put this into Coveralls later.

This PR should be based on the "one mode" PR #50, because it makes this one much easier to develop, but I based it on master for the future. Please disregard the first few commits that are also in #50.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,16 @@
 # working and output directories
 _build/
+_build.*/
+_install.meta/
 ocamldoc/
 
-# test working directory
+# test working and coverage directories
 tests/_scratch
+tests/_coverage
+tests/_report
 
 # generated module lists
-bisect.mlpack
+*.mlpack
 bisect.odocl
 
 # scratch directory for notes, etc.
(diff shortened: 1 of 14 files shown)


rleonid:
wow.

aantron:
Let me rebase this to lose the extra commits.

rleonid:
This is really impressive for what it does and the cleanliness of the implementation. It is also interesting to note the corner cases that are not covered by the test case of `instrumentPpx.ml.` Looks good to merge. I'll give you the honors!

aantron:
Thanks. Going to wait for Travis to do its thing (just to be extra careful), then merge.

PR #52, opened 2016-01-20
Improve HTML legibility

aantron:
- Fonts with clearer characters (font stack, size, line height taken from GitHub, presumably they know what they are doing for each system).
- Desaturated background colors so black text remains visible over them.

Here are some before/afters.

Colors before:

screen shot 2016-01-20 at 00 01 23

Can barely read black text on red when tired. After:

screen shot 2016-01-19 at 23 56 36

Font before:

screen shot 2016-01-20 at 00 04 05

The tildes (in `~loc`) really look a lot like minuses when tired. After:

screen shot 2016-01-20 at 00 05 12

Now clear. Don't know what this looks like on Windows or Linux, but guessing the fonts on those systems are clearer as well.

### DIFF
--- a/src/report/reportHTML.ml
+++ b/src/report/reportHTML.ml
@@ -52,10 +52,17 @@ let css =  [
   "    width: 75%;" ;
   "}" ;
   "" ;
-  ".lineNone { white-space: nowrap; background: white; font-family: monospace; }" ;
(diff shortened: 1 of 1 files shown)


rleonid:
Looks good, merge away.

Issue #35, opened 2016-01-17
Test rewriters in both order.


rleonid:
Bisect_ppx needs to work well with other ppx rewriters in any order:
`ppx_deriving` -> `bisect_ppx`  and `bisect_ppx` -> `ppx_deriving`

aantron:
Do you recall what the difficulties were with this? I saw in the old tests that `ppx_deriving` was only tested one way, and there were also unused files for `ppx_getenv` (IIRC).

rleonid:
I don't immediately remember the difference, I just would make sure that there is a test that allows one to write the rewriters in any order. I'm not certain if the output should be the same.

PR #55, opened 2016-01-21
Rewriter improvements

aantron:
This is a bunch of rewriter fixes (mostly simplifications) and testing improvements. It takes care of #35 and brings coverage of `instrumentPpx.ml` to 100%.

The commit messages/diffs together should give sufficient explanation of almost everything, but here are a couple other notes.

---

The "bare mapping" idea was (IMO) wrong. Its purpose was to prevent points at the beginning of `match` and `function`.

- `match` might not be exhaustive, and the matched expression might raise an exception or fail to terminate. It is useful to know that the `match` was reached, and not just the cases.
- `function` might not be exhaustive, but also the code was preventing generation of points in cases like this:
```ocaml
e1; (* no point here *) function
  | p1 -> ...
  | p2 -> ...
```
Obviously, in this case, the match isn't even being evaluated, just a function is being "returned". It is useful to know that the `function` expression was reached.

---

I deleted special handling for sequences. It was trying to place points at the beginning of each expression in a sequence, and then there was more code to resolve conflicts between the first point in a sequence and `for`.

I think the sequence expression should just result in generation of this point:
```ocaml
e1; (* point *) e2
```
Some other code might place a point on the front of `e1`.


rleonid:
This looks great! It was a bit easier to go commit by commit, but ultimately this is the a good simplification of `instrumentPpx.ml`

aantron:
Yep, I kept the commits separate so they could be reviewed individually. Thanks!

rleonid:
Before
Before

After
After

The difference between the annotations of `f` and `x` in the anonymous function is due to the currying?
This is pretty cool.

aantron:
This difference is from the other PR, right? I should look into it.

Should also give some thought to whether that change will turn some type variables into "weak" type variables (`'a` to `'_a`).

aantron:
I'm having trouble reproducing it with either this PR or the next one.

```ocaml
000050| let blah =
000051|   (*[2]*)let foo = (*[2]*)fun f x -> (*[2]*)f x in
000052|   (*[2]*)let bar = (*[2]*)foo ((+) 1) in
000053|   (*[2]*)bar 2
```

What is the missing ingredient?

rleonid:
Having a messed up install :smile: ! Tried to replicate with latest `master` and I couldn't do it. I'll keep an eye out for it, but I don't think that there is an issue.

PR #56, opened 2016-01-22
Independent points for each clause of or-patterns

### LINKED ISSUE #43: Add points to each clause of or-patterns
Not yet sure if this is reasonable, but I want to look into adding this translation:

```ocaml
match foo with
| `P1 -> e1
| _, (`P2 | `P3 as x) as y -> e2
```

to

```ocaml
let e2' x y = e2 in
match foo with
| `P1 -> point 1; e1
| _, (`P2 as x) as y -> point 2; e2' x y
| _, (`P3 as x) as y -> point 3; e2' x y
```

and reporting as

```ocaml
match foo with
| (* point 1 *) `P1 -> e1
| _, ((* point 2 *) `P2 | (* point 3 *) `P3 as x) as y -> e2
```

In case of nested or-patterns, the point would be on the deepest clauses. In case of a product of or-patterns, I am not sure what to do. There is both a combinatorial blow-up and no good place to put a point. Since Bisect_ppx explicitly doesn't try to understand execution paths, just whether execution passes by a point, this whole thing might be against the spirit of the tool anyway and a bad idea. In that case, I we should briefly mention why we don't do this in some new documentation we may write later.

@rleonid thoughts?

aantron:
Resolves #43.

Here are some before and afters.

I think this is much more useful (after):

screen shot 2016-01-22 at 11 36 32

Than this (before):

screen shot 2016-01-22 at 11 41 56

Also, even for cases without or-patterns, I think the new output is easier to read and make sense of:

screen shot 2016-01-22 at 11 38 42

Before:

screen shot 2016-01-22 at 11 41 24

This took Markup.ml coverage from 78% to 65%, which I think is a good thing. It will encourage me to write much more thorough tests.

Here is a sketch of the translation implemented. If you have

```ocaml
match x with
| (Foo | Bar), (Foo | Bar) -> e
```

Bisect_ppx will now replace it with

```ocaml
match x with
| ((* p1 *) Foo | (* p2 *) Bar),
  ((* p3 *) Foo | (* p4 *) Bar) as ___bisect_matched_value___ ->
  begin match ___bisect_matched_value___ with
  | Foo, Foo -> mark p1; mark p3
  | Foo, Bar -> mark p1; mark p4
  | Bar, Foo -> mark p2; mark p3
  | Bar, Bar -> mark p2; mark p4
  | _ -> ()
  end [@ocaml.warning "-8-11"];
  e
```

**EDIT** Updated above sample.

This PR currently includes all the commits of #55. Only the last commit is unique to here. I will rebase this PR after #55 is merged.


src/syntax/instrumentPpx.ml
```diff
@@ -65,9 +90,13 @@ let marker file ofs kind marked =
    unmodified if the expression is already marked, has a ghost location,
    construct instrumentation is disabled, or a special comments indicates to
    ignore line. *)
```
rleonid: "ghost location, that [kind] of construct instrumentation is ... a special comment indicates" (no 's' in comments)

src/syntax/instrumentPpx.ml
```diff
@@ -87,14 +116,157 @@ let wrap_expr k e =
...
       | Some w -> Exp.sequence ~loc w e
       | None   -> e
 
+(* Given a pattern and a location, transforms the pattern into a list of
+   patterns by transforming or-patterns into cases (i.e., promoting them to the
+   top). Each resulting case pattern is paired with a list of locations to mark
+   should that case be reached. The location argument to this function is the
+   location of the closest enclosing or-pattern clause, or, if there is no such
+   clause, of the entire enclosing pattern. *)
+let translate_pattern =
```
rleonid: "no such clause; the entire enclosing pattern."

src/syntax/instrumentPpx.ml
```diff
@@ -87,14 +116,157 @@ let wrap_expr k e =
     else
       let marked = List.mem line c.CommentsPpx.marked_lines in
       let marker_file = !Location.input_name in
-      match marker marker_file ofs k marked with
+      match marker must_be_unique marker_file ofs k marked with
       | Some w -> Exp.sequence ~loc w e
       | None   -> e
 
+(* Given a pattern and a location, transforms the pattern into a list of
+   patterns by transforming or-patterns into cases (i.e., promoting them to the
```
rleonid: "transform the pattern"

src/syntax/instrumentPpx.ml
aantron: I wanted to keep changes to a minimum. Perhaps we can have a comment cleanup change list at some point?

rleonid: Sure, one commit to clean these up is fine. I was going to do that like last time but I'm a bit pressed for time at the moment.

src/syntax/instrumentPpx.ml
aantron: What's wrong with it? It's just indicative instead of imperative, like some of the older comments in the file.

src/syntax/instrumentPpx.ml
aantron: Confused here, the proposed wording seems ungrammatical.. ?

rleonid:
I was a bit skeptical of this idea but my first pass through convinced me that this looks like a clean implementation and I'm glad to hear that it is helping you find edge cases. I want to do a little bit more due diligence with testing the output since we're rewriting patterns, but I am pretty confident in merging this.

src/syntax/instrumentPpx.ml
rleonid: Ah yes. "location **of** the entire closing pattern" My mistake. I'll try to think of a way to simplify this comment

src/syntax/instrumentPpx.ml
rleonid: There is nothing wrong, just a preference, we can keep it the way it is.

aantron:
Good :) We actually aren't rewriting patterns here. I took your suggestion from the issue, and just added side-effecting `match` expressions to the beginning of each case. We already had analogous expressions in each case before, now they are just more fancy, and based on the case's pattern.

I think the only potential source of errors this introduces into the user's code, relative to the previous scheme, is that the generated `match` expression might fail due to inexhaustive matching, due to a bug in this new Bisect_ppx code. I suppose we could handle that by adding an `_` case to each `match`. That would make the bug harder to find, but it would avoid altering the behavior of the code being instrumented (relative to behavior under previous Bisect_ppx).

src/syntax/instrumentPpx.ml
aantron: I think splitting it into 2 paragraphs might work, but I think I better leave it to you, since you have a fresher pair of eyes :)

aantron:
Tried to fix the comments.

tests/instrument/expr_match.ml.reference
```diff
@@ -99,3 +99,174 @@ let f x =
...
          | Bar  -> (___bisect_mark___expr_match 37; print_endline "barbar"))))
+let f x =
+  ___bisect_mark___expr_match 43;
+  (match x with
+   | Foo |Bar  as ___bisect_matched_value___ ->
+       ((((match ___bisect_matched_value___ with
+           | Foo  -> (___bisect_mark___expr_match 41; ())
+           | Bar  -> (___bisect_mark___expr_match 42; ())
+           | _ -> ()))[@ocaml.warning "-8-11"]);
+        print_endline "foo"))
```
rleonid: Would it be possible to change the recursion so that the expression is just promoted to the appropriate patterns?
```OCaml
let f x =
  ___bisect_mark___expr_match 43;
  (match x with
   | Foo -> (___bisect_mark___expr_match 41; ())
   | Bar -> (___bisect_mark___expr_match 42; ()));
   print_endline "foo"
```
I think that you had a good reason at the time, which I admit, I don't think I fully understand. I think that this might give some 'power' users pause when they look at the expanded, covered source, and it is not entirely obvious why the code isn't as simple as their original.

tests/instrument/expr_match.ml.reference
```diff
@@ -99,3 +99,174 @@ let f x =
...
+let f x =
+  ___bisect_mark___expr_match 93;
+  (match x with
+   | exception (Exn (Foo |Bar ) as ___bisect_matched_value___) ->
+       ((((match ___bisect_matched_value___ with
+           | Exn (Foo ) -> (___bisect_mark___expr_match 90; ())
+           | Exn (Bar ) -> (___bisect_mark___expr_match 91; ())
+           | _ -> ()))[@ocaml.warning "-8-11"]);
+        print_endline "foo")
+   | _ -> (___bisect_mark___expr_match 92; print_endline "bar"))
```
rleonid: Do we also need a 'mixed' pattern that combines an `or` pattern only on one branch:

```OCaml
type g =
  | Cat
  | Dog
  | Eel

let g x =
  match x with
  | Cat | Dog -> print_endline "cat or dog"
  | Eel -> print_endline "eel"

```
```OCaml
let f x =
  ___bisect_mark___expr_match 43;
  (match x with
   | Foo -> (___bisect_mark___expr_match 41; ())
   | Bar -> (___bisect_mark___expr_match 42; ()));
   print_endline "foo"

```

aantron: I am not sure what you mean by a mixed pattern, or combining on one branch. Can you clarify?

rleonid: By a mixed pattern, I just meant a regular case list, where one of the patterns is an or ( `| Cat | Dog ->`) and the other is something else, such as a regular constructor (`| Eel -> `)

tests/instrument/expr_match.ml.reference
aantron: Yes, this was my first instinct. The main reason against doing "full promotion" is that if there is a bug in Bisect_ppx's pattern rewriter, it can mess up the meaning of the user's code. That could silently make their testing misleading or useless.

With nested match expressions like now, the bug might mess up coverage statistics, but it should not harm the correctness of code under test. There are other, lamer, reasons, such as it is easier to wrap the pattern with `as ___bisect_matched_value___` than to move the case body out to a let-expression, find all its free variables, turn it into a function, and then apply it. That's what I think we would have to do, and, by the way, it might also look strange, at first, if someone examined the output.

As for power users, I think we will have to be clear that the advantage of repeated matching, as proposed now, is correctness of testing, and that the disadvantage only applies during testing and has nothing to do with their production code anyway. Not sure how to go about this, maybe a FAQ will be needed, or maybe it will be a non-problem anyway.

In any case, the algorithm I have now, and the algorithm you are asking about, are near relatives of each other, and I wrote the `transform_pattern` function in such a way that it can be used to do what you want instead. It should be only slightly annoying to change between the two, if it turns out to be necessary or desirable.

Let me know your thoughts.

tests/instrument/expr_match.ml.reference
aantron: I am not sure what we need that for. I think I am missing some context for why we would need to generate this.

tests/instrument/expr_match.ml.reference
aantron: Well, we wouldn't have to find free variables, we would have to find the variables bound in the pattern.. anyway, it's slightly involved.

aantron: I think there are actually some other problems, e.g. let's say `s : unit -> bool` is a side-effecting function and you have

```ocaml
match v with
| Foo | Bar when s () -> ...
| _ -> ...
```

If you rewrite this to e.g.

```ocaml
match v with
| Foo when s () -> ...
| Bar when s () -> ...
| _ -> ...
```

You will cause the side effect twice. It's getting late so I am not sure if there is something else you could obviously rewrite this to, that wouldn't have such a problem, but I don't think so right now.

rleonid: I think you meant that we'll cause the side effect twice if _both_ patterns match; but I think that is exactly what we want. My understanding of `or` patterns in `OCaml` is that there should not be a difference between your two previous examples, and that the first form is just a shorter (more convenient) way for of writing the second.

With regard to promoting expressions, again, I'm not so certain that this would be that difficult. As I understand it, both forms of the expression would receive the same "closure" (in quotes because I'm not certain I'm being precise; I mean the set of all bound variables), so no extra search is necessary.

I completely agree that the way you implemented `translate` this should be easy. I unfortunately, don't have the time to dive in and try implementing that approach. I've tested this enough to feel confident that this works so we can merge, but I'd like 3 things:

1. can we add a command line switch to disable this "or" expansion ? I think that this is a powerful tool and should be on by default, but a switch might help to easily diagnose weird corner cases that we might miss.
2. can you also add the above explanation to the docs (README.md), it is pretty good.
3. I want to file an issue, for myself, to write the "or" expansion with expression promotion. It will help me to understand the problem better, and we can not merge if there is no significant improvement.

tests/instrument/expr_match.ml.reference
rleonid: You're right this is not needed. I saw this as a case requiring two levels of recursion, while your previous examples (as far as I understood) were all one level. But because of the coverage of our coverage tool, you had all the corner cases tested!

tests/instrument/expr_match.ml.reference
aantron: 1. Very well, I will add the switch. Seems like a good idea.
2. As for the docs, I will of course mention what we do with or patterns. Which part of the explanation are you referring to for `README.md`? We have a lot of text above :) I will have a PR or PRs soon with documentation drafts, so we should be able to discuss it in more detail there. If you don't mind, I will make a note of it now, but modify the docs in that later PR.
3. Sounds good as well :)

My example with the pattern guard above was bogus, because `s ()` won't be run anyway in the second `match` expression unless `Foo` or `Bar` match first, and only one of those can match.

However, here is a slightly ridiculous example that *does* have different behavior between the two expressions. I tested in the top-level to make sure I haven't made a mistake again:

```ocaml
let s () = print_endline "foo"; false;;

match 0 with
| _ | _ when s () -> ()
| _ -> ();;

match 0 with
| _ when s () -> ()
| _ when s () -> ()
| _ -> ();;
```

The first prints `foo` once, the second twice. Granted, the first gives a warning (unused sub-pattern), but a `match` expression with redundant sub-patterns might conceivably result automatically from another preprocessor, and not be inherently suspicious.

tests/instrument/expr_match.ml.reference
aantron: As a little coverage success story, Bisect_ppx code coverage output helped me catch several pattern translation bugs before I uploaded this, by just making me think harder :)

aantron:
Added the flag. Tested manually by changing the flag's default to `true`, running the unit tests, and examining the diffs from the resulting failures.

tests/instrument/expr_match.ml.reference
rleonid: This is a good counterexample, that I can think can lead to better ones. It weirdly reminds of how side effect expressions are not allowed in `Erlang` guards, which at the time I thought was impure. 

I'd include the first 2 paragraphs of your explanation in the doc.

rleonid:
Looks good to merge.

PR #57, opened 2016-01-22
Support writing to STDOUT for -bisect output

### LINKED ISSUE #33: -bisect report output mode does not support writing to STDOUT when filename is '-'
All other modes (besides HTML, which creates a directory), do.

https://github.com/rleonid/bisect_ppx/blob/530a713b4227e3b55a4d8a0c5f1a28ca25df9a6c/src/report/report.ml#L91-L104

aantron:
Tested by temporarily changing report `-bisect` test case to take output from `STDOUT`. No permanent test included.

Fixes #33.


rleonid:
Looks good.

PR #58, opened 2016-01-24
Update appearance of HTML reports

aantron:
See the commit message. I'm also thinking about adding optional syntax highlighting, when a highlighter is available.

This PR is more a proposal than something complete. Would be very interested to hear/see how you think the reports should look.

---

**Before**:

old-html

---
---
---

**After**:

new-html

### DIFF
--- a/src/report/report.ml
+++ b/src/report/report.ml
@@ -78,7 +78,6 @@ let main () =
         mkdirs dir;
         ReportHTML.output verbose dir
           !ReportArgs.tab_size !ReportArgs.title
-          !ReportArgs.no_navbar !ReportArgs.no_folding
           search_in_path data
     | ReportArgs.Xml_output file ->
         generic_output file (ReportXML.make ())
--- a/src/report/reportArgs.ml
+++ b/src/report/reportArgs.ml
@@ -34,13 +34,9 @@ let verbose = ref false
 
 let tab_size = ref 8
 
-let title = ref "Bisect report"
+let title = ref "Coverage report"
 
 let separator = ref ";"
-
-let no_navbar = ref false
-
-let no_folding = ref false
 
 let search_path = ref [""]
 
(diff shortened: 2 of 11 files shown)


c-cube:
This looks nice!

rleonid:
This is looks great; no reservations about making the switch or proceeding with these choices.

I noticed a turquoise color for a line, how is that different from a yellow?

aantron:
That's a line that got clicked on. It's a direct link to that line (`#L79`), and the line is highlighted in some garish color meant to stand out from the others. I'm not necessarily happy about the specific color, but I gave myself finite time to tweak them, and ran out. If you have any suggestions, I will be happy to change it, or go ahead and do so yourself :)

aantron:
Perhaps a little CSS animation is a better choice both to attract attention to the clicked line, and to disappear and not be annoying afterwards, once the attention is on it... Let me try.

aantron:
What about this? I also took the opportunity to make the navbar lines a lot fatter. They were difficult to click on.

navhighlight

rleonid:
Haven't tested, but looks good. To the extent possible with the CSS animation, would it be easy to 'center' to that line when the navbar line is clicked?

aantron:
I don't think it's possible with CSS.

But we already have a bit of JS here, so I'll experiment with a script that
- Scrolls the viewport when the navbar is clicked, as you suggest, but aligns the line so that it is adjacent to where the mouse pointer was (if possible). I'll compare it to centering.
- Doesn't scroll the viewport when clicking on a line number, but highlights it briefly to hint that the URL has changed. The current scrolling is a bit annoying in this case.

I was also thinking about just making the whole navbar clickable using JS. Then, it doesn't matter how fat the indicators are, as long as they are visible, since there won't be any problem with trying to hit them.

aantron:
Now like this:

highlight2

rleonid:
Looks good.

PR #59, opened 2016-01-25
Packaging and organization changes and fixes

### LINKED ISSUE #54: Fix packaging
I'm noting this for later, because it's probably best done after the "big" changes are in, but:

- Using `git describe` to generate `version.ml` probably means that our code can't be built without a git repo, e.g. if downloaded by OPAM from the GH releases link. We need another way of generating this file, but I would also suggest just hardcoding it (it's no worse than having `VERSION` like before), and adopting a habit of doing a `grep` for the old version number when bumping versions.
- Need to edit the tests section in `opam` to do `make dev` before `make tests`.
- I don't think the `str` dependency in `META` is necessary, since `str` gets linked into the binaries but is not needed for the runtime. Haven't tested.

aantron:
Fixes #54. See the commit message.

I want to keep this PR around for a while, until it's time to increment the version or so. I want to put the changes up for discussion in the meantime.


rleonid:
All of the reorganization changes and new documentations look good. I guess we'll do something "cleverer" for the version at a later point. I'd be fine with jettisoning it from the source entirely and just keeping track via `opam` and `git` tags.

aantron:
Well, re. jettisoning it, `Version` is used by the reporter, both for the `-version` flag, and inlined into the output in XML and HTML. Hopefully the note to `grep` for it in the release checklist will help keep it in sync.

rleonid:
Sounds good.

PR #61, opened 2016-01-25
Mark |> expressions as sequences

aantron:
This adds points like:

```ocaml
l |> (* pt *) List.iter (fun i ->
  i
  |> (* pt *) string_of_int
  |> (* pt *) print_endline);
```

by analogy with sequence expressions. The points belong to kind `Sequence`. We already have special-case handling for `||` and `&&`, so it should not seem too strange to handle `|>` specially.


rleonid:
:+1:

PR #62, opened 2016-01-26
Adjust assignment of points to kinds

aantron:
See the commit message.


rleonid:
That sounds reasonable.

PR #64, opened 2016-01-27
Suppress warning 27 in or-expanded instrument

rleonid:
The instrumented code was failing because this
```OCaml
  match bisection_explicit ~epsilon ~lower ~upper f with
  | `EqZero v 
  | `CloseEnough v -> v
  | `Outside _     ->
```
was getting expanded to
```OCaml
   | `EqZero v|`CloseEnough v as ___bisect_matched_value___ ->
       ((((match ___bisect_matched_value___ with
           | `EqZero v -> (___bisect_mark___src/lib/solvers 64; ())
           | `CloseEnough v -> (___bisect_mark___src/lib/solvers 65; ())
           | _ -> ()))[@ocaml.warning "-8-11"]);
        v)
``` 
and the `v` in the inner match is unused.  This seems easier than trying to do something like `v` -> `_v` robustly.

### DIFF
--- a/src/syntax/instrumentPpx.ml
+++ b/src/syntax/instrumentPpx.ml
@@ -277,7 +277,7 @@ let wrap_case k case =
          never be exhaustive. *)
       let marks_expr =
         Exp.attr marks_expr
-          (Location.mkloc "ocaml.warning" loc, PStr [Str.eval (strconst "-8-11")])
+          (Location.mkloc "ocaml.warning" loc, PStr [Str.eval (strconst "-8-11-27")])
       in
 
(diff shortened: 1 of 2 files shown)


rleonid:
My apologies for the poor branch name, should have been `add_unusued_var_warning`

aantron:
LGTM with squash. Noticed this too. BTW, I another way to solve this is to translate variable patterns to wildcards, since we generate no guards, but I think this is the better way.

rleonid:
Thanks for holding me to a higher standard.

aantron:
It's only because (in this case) I'd recommend not committing to `master` an intermediate change that we know breaks testing. It might complicate history investigations in the future.

PR #66, opened 2016-01-28
Test installation and doc generation in Travis

aantron:
To make it less likely that we will break these things without noticing.


rleonid:
:+1:

PR #67, opened 2016-01-28
Support for bytecode-only architectures

### LINKED ISSUE #63: Support bytecode architectures
Our `.itarget` file causes us to build the native reporter unconditionally. We may want to relax that in light of this thread https://sympa.inria.fr/sympa/arc/caml-list/2016-01/msg00038.html

aantron:
See the commit message. Fixes #63.


rleonid:
How would we test this? Just briefly browsing `opam switch --help` I didn't see anything about install just the _byte_ or _native_ versions. I don't have access to any such machines, and trying to get some kind of cloud instance seems overkill. Any idea's ?

aantron:
Well, with respect to just our build process, in Travis I used `BYTECODE_ONLY=yes` to call only `ocamlc`. I put an `ls` after `make build` so we can see which files actually got built in `src/syntax` (as a sample directory) – see, e.g., https://travis-ci.org/rleonid/bisect_ppx/jobs/105453320#L316. This is later followed by an `opam install` with the same environment variable, so I guess the installation works. Of course, I might have missed something, so a "real" test would be better.

Actually, there is a mistake, because I am still building `cmxa` files. Let me fix that.

The test suite still compiles both native and bytecode, but testing is not necessary for installation.

As for detection, I am looking for information. But, in the worst case, the user supplies `BYTECODE_ONLY=yes` and it "should" work. Hopefully they would also submit a bug report and tell us how to detect bytecode-only correctly or what kind of machine to test on.

aantron:
As well as the typo in the commit brief.. urgh.

aantron:
I haven't yet found how (if it's possible) to get a bytecode-only switch. However, another thing we could do to make our testing more likely to be accurate is mess with the findlib configuration in Travis so that `ocamlopt` becomes a command that exits with 2 right away.

rleonid:
> is mess with the findlib configuration in Travis so that ocamlopt becomes a command that exits with 2 right away.

That sounds easy :smile:. In general this looks fine, feel free to merge when ready.

aantron:
Ok, going to do a few investigation builds in Travis first to see if I can override `ocamlopt` easily.

aantron:
Ok, Bisect_ppx now builds with shadowed `ocamlopt`. The test suite also works.

I still don't know if the detection code for a bytecode-only architecture is correct. However, with this change, we should only have to update the detection code, and not the whole build system, if someone later finds a problem.

PR #69, opened 2016-01-28
Ocamlbuild plugin helper module

aantron:
Added a module for use in Ocamlbuild plugins, which should make it easy to

1. Decide which files should get instrumented (tag `coverage`).
2. Build those files both for testing (`BISECT_COVERAGE=YES`) and release (no instrumentation generated).

I will describe how to use the plugin in `advanced.md`, when discussing Ocamlbuild. With this plugin, it should only take a couple sentences.

Also added a usage test, which runs in Travis after Bisect_ppx is installed by OPAM. This test makes sure that the installed rewriter, runtime, reporter, and plugin module are all usable by third-party programs.

This pull request is currently based on #67, because #67 makes this one easier to develop. I will rebase this one when #67 request is in, in some form.

### DIFF
--- a/src/ocamlbuild/bisect_ppx_plugin.ml
+++ b/src/ocamlbuild/bisect_ppx_plugin.ml
@@ -1,0 +1,35 @@
+(*
+ * This file is part of Bisect_ppx.
+ * Copyright (C) 2016 Anton Bachin.
+ *
+ * Bisect is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Bisect is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


src/ocamlbuild/bisect_ppx_plugin.ml
```diff
@@ -0,0 +1,35 @@
...
+let _environment_variable = "BISECT_COVERAGE"
+let _enable = "YES"
+
+let handle_coverage () =
+  if getenv ~default:"" _environment_variable <> _enable then
+    mark_tag_used _tag_name
+  else begin
+    flag ["ocaml"; "compile"; _tag_name] (S [A "-package"; A "bisect_ppx"]);
+    flag ["ocaml"; "link"; _tag_name] (S [A "-package"; A "bisect_ppx"])
+  end
```
rleonid: Does this require "-use-ocamlfind" argument to `ocamlbuild`?

aantron: Yes, I believe so. I should note it in the documentation.

rleonid:
Looks fine, merge when ready.

PR #71, opened 2016-01-30
Set up GitHub Pages

aantron:
This adds a `Makefile` target `gh-pages` for generating and publishing the GH Pages page for Bisect_ppx. See the [home page](http://rleonid.github.io/bisect_ppx/) and live [coverage report](http://rleonid.github.io/bisect_ppx/coverage/) – that's the entire site.

Doing this requires [OMD](https://github.com/ocaml/omd) and [Lambda Soup](https://github.com/aantron/lambda-soup). Since only repo collaborators can run this target, these aren't dependencies of Bisect_ppx.

@rleonid Can you please set the URL in the GitHub header to point to http://rleonid.github.io/bisect_ppx/coverage/? We could also point it to the home page, but that page seems redundant with `README.md`. Also, I think we should change the description to something like "Code coverage for OCaml."

I guess we could get fancy and postprocess every page in the live coverage report to point back to GitHub, among other modifications. What do you think about that? I think it's fine as is, but I'd be curious to hear opinions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ _build/
 _build.*/
 _findlib/
 ocamldoc/
+gh-pages/
 
 # test working and coverage directories
 tests/_scratch
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,23 @@ doc: FORCE
 	mkdir -p ocamldoc
 	cp _build/bisect.docdir/*.html _build/bisect.docdir/*.css ocamldoc
 
+GH_PAGES := gh-pages
+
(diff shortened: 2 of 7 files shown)


rleonid:
This looks great, thanks again.

> @rleonid Can you please set the URL in the GitHub header to point to http://rleonid.github.io/bisect_ppx/coverage/

Is this next to the description or is this something else? I do think it is fine the way it is.

aantron:
Yeah, next to the description.

aantron:
Thanks!

PR #73, opened 2016-01-30
Submit coverage reports to Coveralls

aantron:
[Now in Coveralls](https://coveralls.io/github/rleonid/bisect_ppx). I will add badges later. I usually use shields.io for consistent appearance, but this service became unreliable recently. I will wait a bit to see if there is any word on what the problem is, and decide then whether to use direct badges from Coveralls, Travis, etc.


aantron:
Should also mention that for now we install the original Bisect in Travis as a dependency of ocveralls :) Will fix this later. There seems to be some kind of issue when installing with ocveralls with Bisect_ppx.

aantron:
After more reflection, I think it's better to put this PR on hold until I figure out what's happening.

aantron:
https://github.com/sagotch/ocveralls/pull/12 should fix it. After that, this will double as a nice integration test of installed Bisect_ppx + ocveralls.

PR #75, opened 2016-02-05
Bump version to 1.0.0

aantron:
Changelog up for review; to be merged after the other PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Bisect_ppx &nbsp; [![version 0.2.6][version]][releases] [![Travis status][travis-img]][travis]
+# Bisect_ppx &nbsp; [![version 1.0.0][version]][releases] [![Travis status][travis-img]][travis]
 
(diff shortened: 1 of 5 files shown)


rleonid:
Excellent, :+1:

aantron:
This doesn't have a release date in the changelog, since we are "soft" blocked on ocveralls. I wanted to merge it after #73. `git push -f`?

rleonid:
Woops. Yea, let's do a git push when that is unblocked. We're waiting for `ocveralls` to be updated in `opam`?

aantron:
Yes. It's not necessary, but it would be nice, since then we could be sure that our packaging in 1.0.0 is compatible with one released version of ocveralls. As of now, the only way to use any released ocveralls is with the original Bisect or with Bisect_ppx <= 0.2.6. https://github.com/sagotch/ocveralls/pull/12

Do you think it's better to just go ahead?

rleonid:
I don't mind waiting, gives us time to prepare.

PR #77, opened 2016-02-07
Documentation nits

aantron:


rleonid:
In the part where you explain how to use the `ocamlbuild` plugin, I could imagine a novice still being confused as to why one needs the `-plugin-tag 'package(bisect_ppx.plugin)'` perhaps adding a line such as: "Or add `mark_tag_used "coverage"`into your `myocamlbuild.ml` " . I could also imagine a strict GPL enforcer type person not getting it. What do you think?

rleonid:
Read the rest of the documentation and this looks great!

aantron:
- Regarding the tag, what (else) do you think it should say?
- Just doing `mark_tag_used "coverage"` will not result in coverage reports. If you don't also add the plugin, you will end up with reports unconditionally off, so I think the plugin is required.
- I am not sure what you mean by strict GPL enforcer not getting it – like, people will be afraid to use this because of GPL? I can see how that could be a problem, so perhaps we should make the disclaimer bigger, or we can put the plugin in the public domain and make that clear. That shouldn't be a problem since we created it, so there is no previous license to mind.

rleonid:
Sorry, I wasn't being clear. The part where we write
```
# For release
  ocamlbuild -use-ocamlfind \
          -plugin-tag 'package(bisect_ppx.plugin)' src/my_program.native
```
a novice or a 'strict GPL enforcer' might object to since we're still referencing something bisect_ppx related, ie `-plugin-tag 'package(bisect_ppx.plugin)'`. We could instead instruct them to modify the `myocamlbuild.ml` to use `mark_tag_used "coverage"` and build as usual for when they want to release.

aantron:
Perhaps I still am not understanding, but I think we shouldn't tell people to modify `myocamlbuild.ml` just to switch from building for testing to building for release. That seems like something annoying to do repeatedly.

rleonid:
I agree that it is annoying. Perhaps we can release the plugin with LGPL license or something a bit more permissive than GPL?

aantron:
I think that's a good idea. I would suggest something even more permissive though, such as public domain. The plugin is trivial and we don't really care if anyone modifies it or incorporates it into their code without attribution. We want that to be easy and stress-free.

rleonid:
:+1:

aantron:
Ok, perhaps we should include this in the release. Here is the new wording for the Ocamlbuild section: https://github.com/rleonid/bisect_ppx/blob/doc-fixes/doc/advanced.md#Ocamlbuild

And a sample license header: https://github.com/rleonid/bisect_ppx/blob/doc-fixes/src/ocamlbuild/bisect_ppx_plugin.mli

rleonid:
Awesome, thank you for addressing my paranoia.

PR #80, opened 2016-02-09
Add projects that use Bisect_ppx

aantron:
I did a GitHub search for Bisect_ppx. I tried to pick "well-known" projects, and took only those that post a coverage report that we can link to. However, I am not in a position to fairly state if Markup.ml qualifies as "well-known." Indeed, it's quite young and unproven. I can, however, say that it has a lot of code, and benefitted greatly from Bisect_ppx. It brought me here :)

In general, due to our involvement, I would prefer to list neither Oml nor Markup.ml in the project list, under such a subjective criterion. However, this would leave us with only one (or two, if we keep Oml) project in the list.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,8 @@ use it for tracing: run one test, and see what is visited.
 <br>
 
 For a live demonstration, see the [coverage report][self-coverage] Bisect_ppx
-generates for itself.
+generates for itself. You may also want to see
+[projects that use Bisect_ppx](#bisect_ppx-in-practice).
 
 [self]:          https://github.com/rleonid/bisect_ppx
 [releases]:      https://github.com/rleonid/bisect_ppx/releases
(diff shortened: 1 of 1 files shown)


aantron:
Meant to also say: I would have liked to mention containers and ocamllint, but they don't (apparently) post reports.

aantron:
And, is it Oml which caused you to fork Bisect?

rleonid:
I think that we can go without those tools that report coverage. For example [Ketrew](https://github.com/hammerlab/ketrew/blob/master/OMakeroot#L24) and [Sosa](https://github.com/hammerlab/sosa/blob/master/Makefile#L26) use bisect_ppx. It was Ketrew that got me to fork it.

aantron:
Without, as in beyond? Could you take over this PR from here?

rleonid:
Sorry, (just can't write well today). I meant to say that we can list those projects without pointing to their coverage results.

rleonid:
Yes, I'll take over and add some, just a sec.

aantron:
The main reason I wanted to restrict the list to projects that post reports is that I wanted it to be a list of demos useful to visitors, more than a list showing adoption of Bisect_ppx. I don't have a strong opinion on this, however.

rleonid:
Fair enough, then I won't search for more projects. Perhaps as part of the release to the mailing message, we can add "If you're using Bisect_ppx for your work please let us know."

rleonid:
Merge this when ready.

aantron:
Ok, I will work an invitation into the announcement. By the way, did you configure Coveralls to be really strict about what counts as a failure? Right now, we are getting failed checks for no coverage changes.

Issue #86, opened 2016-03-03
Doesn't work well with `findlib.dynload`


ivg:
The `findlib.dynload` library is used by a host program to load plugins. It tracks already loaded modules, so that we can prevent corruption. It does this by embedding special initialization code into the binary. The code is put into the temporary file, and the file is later deleted. When we run `bisect-ppx-report` on the output of such binary, we get an error:
```
 *** system error: /tmp/findlib_initl44dd7c.ml: No such file or directory
```

Maybe it is possible to add some command-line flag, that will allow to ignore missing files? Or just relax this error to a warning?

PR #87, opened 2016-03-03
Add option for reporter to ignore missing files

aantron:
This adds a new flag to `bisect-ppx-report`, per #86.

    -ignore-missing-files  Do not fail if an .ml file can't be found

I tested it locally:

    ocamlfind opt -package bisect_ppx -package findlib.dynload -linkpkg foo/foo.ml && \
    ./a.out && \
    bisect-ppx-report -ignore-missing-files -html report bisect*.out

Works with this flag and doesn't work without it. Will amend with automatic tests shortly.

@ivg, is it convenient enough for your project to use a dev verison of Bisect_ppx, or do you prefer a speedy release to OPAM?


ivg:
@aantron, it is totally ok for us to use the dev version, there is no need for a rush.

Thanks a lot for a speedy response!

aantron:
No problem. Please let me know if this fix works for you, at your leisure :)

ivg:
Looks like that it works. But I'm not sure that I'm using it correctly, as my report consists only from one file. Here is the output of the `bisect-ppx-report`
```
$ bisect-ppx-report -ignore-missing-files -verbose -I _build -I lib/bap_types/ -html coverage/ bisect*.out
Processing file '/tmp/findlib_initl231fa4.ml'...
... file not found
Processing file 'lib_test/bap/run_tests.ml'...
... file has 14 points
Writing index file...
```

Now, how I'm compiling. I'm using oasis, but I didn't follow your instructions ... basically, because our build system doesn't require it (I maybe wrong of course). Here is the command,  that is used to build the test file:

```
ocamlfind ocamlopt -linkpkg -g  -linkpkg -package bisect_ppx -package <other-packages> lib/bap_bundle/bundle.cmxa lib/<all-other-cmxa> lib_test/bap/run_tests.cmx -o lib_test/bap/run_tests.native
```
It looks for me, that this should be enough, am I right?

aantron:
This looks like the linking step, and it looks fine. In the compilation steps, are you compiling only `run_tests.ml` with the `bisect_ppx` package? If so, that would explain why only `run_tests.ml` is in the report.

ivg:
aha! now it makes sense!

ivg:
finally it compiles, so we've covered much more files, but failed with a drastic `Not_found`. The log is attached. Meantime I will try to get a backtrace

[bisect.txt](https://github.com/rleonid/bisect_ppx/files/158907/bisect.txt)

ivg:
and here comes a backtrace:
```
Fatal error: exception Not_found
Raised at file "buffer.ml", line 123, characters 27-36
Called from file "buffer.ml", line 145, characters 16-62
Called from file "buffer.ml", line 164, characters 29-47
Called from file "src/report/reportUtils.ml", line 100, characters 6-38
Called from file "list.ml", line 73, characters 12-15
Called from file "list.ml", line 73, characters 12-15
Re-raised at file "src/report/reportHTML.ml", line 701, characters 12-13
Called from file "src/report/reportHTML.ml", line 716, characters 10-123
Called from file "hashtbl.ml", line 200, characters 23-35
Called from file "hashtbl.ml", line 204, characters 12-33
Called from file "src/report/reportHTML.ml", line 708, characters 14-636
Called from file "list.ml", line 73, characters 12-15
Called from file "src/report/report.ml", line 110, characters 4-11
```

aantron:
Wow. Ok, thank you. Taking a look.

aantron:
Is it possible to see the apparently generated file `lib/bap_piqi/stmt_piqi.ml`?

ivg:
Thank you. 
Of course it is possible (it is indeed nasty btw)

[stmt_piqi.zip](https://github.com/rleonid/bisect_ppx/files/158937/stmt_piqi.zip)


Btw, the `buffer.ml` in mine OCaml installation points to:
```ocaml
(* opening and closing: open and close characters, typically ( and )
   k: balance of opening and closing chars
   s: the string where we are searching
   start: the index where we start the search. *)
let advance_to_closing opening closing k s start =
  let rec advance k i lim =
    if i >= lim then raise Not_found else <----------- HERE
    if s.[i] = opening then advance (k + 1) (i + 1) lim else
    if s.[i] = closing then
      if k = 0 then i else advance (k - 1) (i + 1) lim
    else advance k (i + 1) lim in
  advance k start (String.length s);;
```

So it looks like an unbalanced parentheses issue in the substitution.

aantron:
Yes, thank you. Actually, I think I already found the problem. Will upload the patch in some minutes.

aantron:
It was a classic arbitrary-input-as-format-string bug. Should be fixed now – try the commit just posted.

ivg:
Thanks! Will try. Meantime I've tried with piqi disabled, and indeed, you were right, it was the cause of problem. Without the piqi, everything works fine :)

ivg:
Checked with the piqi-enabled. Everything works fine, and the piqi file is included in the report. Thanks alot!

aantron:
Thanks for the reports. Given that we fixed a correctness bug, this should end up in OPAM pretty soon.

aantron:
Unfortunately still not in OPAM due to whatever is causing https://github.com/ocaml/opam-repository/issues/5899.

Issue #89, opened 2016-03-28
Support regex in filename exclusions


solongordon:
Currently it looks like although top-level values can be excluded by regex, excluded filenames must be specified exactly in the exclusion file. It would be great to support regex for filenames as well, for example to allow excluding unit tests themselves from coverage via `^test_.*`.

aantron:
@solongordon If you are using Ocamlbuild and the plugin, you should be able to achieve this effect with a line like

```
<**/test_*>: -coverage
```

Let us know if that doesn't help.

aantron:
(Note) without the plugin, you should be able to do `-package(bisect_ppx)` instead.

solongordon:
Thanks for the suggestion. We're using OMake so unfortunately we can't use the _tags approach.

We can certainly get by with the currently exclusion file format, and/or by only building some files with bisect_ppx enabled, but it would be more convenient to exclude via glob or regex.

rgrinberg:
Just a heads up, excluding all the test files using ppx_bisect in omake was easy enough but we found skipping out on some other generated files to be easier done with ppx_bisect's config file.

We thought it would just be better if all of those settings lived in 1 place.

PS: I work with @solongordon

aantron:
Just to be clear: by config file you mean the exclusion file (for option `-exclude_file`)?

solongordon:
Yup, exactly.

aantron:
Are you okay with using the dev version of Bisect_ppx for now? The release to OPAM is blocked on some 4.03-related questions for a bit. If you would prefer the updated Bisect_ppx be available in OPAM rather than using a dev version, I can make a release including this change but without 4.03 compatibility.

solongordon:
Yes, no rush on the OPAM release. Thanks again.

PR #90, opened 2016-04-01
Allow exclusion of entire files; files by regex.

aantron:
@solongordon @rgrinberg this change gives syntax as below, for excluding specific whole files, and whole files by pattern:

```
file "specific_test.ml"
file regexp "test_.*"
```

Files excluded in this way don't get instrumented at all, and don't appear in the coverage report. You can also combine the file regexp with the old top-level value exclusions to do things like

```
file regexp "src/.*" [ regexp "test_" ]
```

to exclude all top-level values beginning with `test_` in all files in `src/`.

Let me know if this addresses #89 sufficiently. If so, will adjust documentation and merge in.


solongordon:
This looks great! Exactly the functionality we were looking for. Thanks for addressing the feature request.

rgrinberg:
:100: Don't forget to mention it's Str's regex syntax.

aantron:
Sorry about the coveralls spam. :(

aantron:
This is now installable from OPAM as `bisect_ppx` 1.1.0.

PR #106, opened 2016-06-04
Reporter cleanup

### LINKED ISSUE #105: Should be deprecated and/or removed from reporter
1. Remove EMMA reports.

  > This presumably tries to mimic [EMMA](http://emma.sourceforge.net/), an old Java coverage tool. Wikipedia [states](https://en.wikipedia.org/wiki/Java_Code_Coverage_Tools#EMMA) that it has not been maintained since 2005 or so. The presumed replacement JaCoCo seems to have incompatible XML reports ([JaCoCo DTD](http://eclemma.org/jacoco/trunk/coverage/report.dtd) vs. [our EMMA code](https://github.com/aantron/bisect_ppx/blob/f3be018558e7801a91678b04b07527f8eaf32b78/src/report/reportXMLEmma.ml)).

2. Remove XML, CSV, dump reports, "bisect" reports.

  > None of these are used much (no uses in GitHub). We should pick one machine-friendly output format, and provide only that. @lindig suggested JSON – that may be the best idea. Outputting coveralls.io JSON does not seem good, however, because that loses information about points and measures only lines. @sagotch?
  >
  > EDIT: Actually, I think `-text` is going to be quite easy to read, once we get rid of point kinds. So, maybe we can remove all of these. If someone later has a strong reason for supporting an additional machine-friendly format, we can consider it then.

3. De-emphasize `-combine-expr`.

  > This seems like a cool feature in principle, but I don't think anybody is using it (0 uses in GitHub between Bisect_ppx and the original Bisect).

aantron:
Resolves #105.

`-no-navbar` and `-no-folding` were deprecated, but I doubt anyone ever used them anyway. They were used in the tests at some point, when the old HTML reports were around.

I kept the "dump" output format for now, because it useful for the thread safety test.


rleonid:
I suggest keeping the CSV format, it at least gives some guarantees about structure over the regular text format.

aantron:
Fair enough. I've omitted that commit.

aantron:
Got no objections ([1](https://sympa.inria.fr/sympa/arc/caml-list/2016-06/msg00030.html), [2](https://sympa.inria.fr/sympa/arc/caml-list/2016-06/msg00004.html), [3](https://www.reddit.com/r/ocaml/comments/4ngdtv/bisect_ppx_any_objections_to_removing_these/)), so merging this and #107.

Issue #100, opened 2016-05-29
What is the code point for runtime.(i)?


lindig:
I have a question about how to find the code point for an entry in the runtime data. Given for a source file a counter in its array `runtime` at index `i` (that is, `runtime.(i)`), what is the corresponding code point? Is it the one with identifier `i` or the one at index `i` in the list of code points? 

I'm not sure about the purpose of the `identifier` field in type `point_definition` as every point definition could be identified by its position in the list of all code points.

```
type point_definition = {
    offset : int; (** Point offset, relative to file beginning. *)
    identifier : int; (** Point identifier, unique in file. *)
    kind : point_kind; (** Point kind, used for statistics. *)
}
```

I think this connection could be documented by also providing a type definition for the runtime data and documenting it in `common.mli`.

aantron:
It looks like `identifier` is simply the value of the point's index `i`. It's just a way to get the index of a code point, given only the `point_definition` record, without doing a search in the points list. So, the answer to

> what is the corresponding code point? Is it the one with identifier i or the one at index i in the list of code points?

is "both."

The naming of the field is confusing though, and the comment is not very helpful. We should fix that.

As for the type definitions, I agree it would make it clearer to understand what is really going on. Both the runtime data and the code point lists (i.e. `.cmp` files) could use clear type definitions. I'll submit a PR – unless you would like to :)

lindig:
Thanks for the clarification. I haven't looked at this but I would try to get rid of this redundancy by removing the `identifier` field as there is the danger of inconsistencies. I'm currently working towards a more abstract interface to the runtime system but it has some parts that you are probably not going to be interested in: I want line numbers for each point. It looks like this and does not rely on array indices as they entire structure is tree shaped.

    module SM: Map.S with type key = string

    exception Error of string

    type point = (** a code point in a source file *)
      { line:         int           (** line number *)
      ; offset:       int           (** byte offset from start *)
      ; kind:         Bisect.Common.point_kind  (** kind of point *)
      ; count:        int           (** number of hits *)
      }

    type file = (** a file with coverage analysis *)
      { digest:       string        (** cryptographic hash of content *)
      ; linecount:    int           (** number of lines *)
      ; points:       point array   (** points that were covered *)
      }

    type t = file SM.t              (** a collection of files *)

    val read: string -> string list -> t

The `read` function takes the prefix directory and a list of `*.out` files and buds a map that for each file has all the points and the number of hits.

aantron:
[...]

PR #108, opened 2016-06-15
Eliminate .cmp files

### LINKED ISSUE #102: Eliminate .cmp files
Christian Lindig (@lindig) has suggested providing a report mode that doesn't require reading the `.cmp` files, which would be useful in testing environments where they are not readily available (note: see [lindig/bisect-summary](https://github.com/lindig/bisect-summary)). In fact, `bisect-ppx-report -text - -summary-only` doesn't actually need the `.cmp` (or source) files, but it reads them anyway.

I think it's a good idea to do this at least for `-text -summary-only`, and probably `-text`, in any case, but we should remove `.cmp` files entirely.

- A `.cmp` file is basically a point list: for each point, its offset into the file and kind. We could inline this information at the top level of each instrumented file, and write it into `.out` files. `bisect-ppx-report` would then deduplicate this information when reading those files.
- I never was fully comfortable with the instrumenter writing `.cmp` files next to the source files as a side effect of operation. That works out fine under some build systems like Ocamlbuild, because the build is done on copies of source in `_build`, but it seems messy otherwise (even if only as messy as having `.cm[iox]` files everywhere).
- It should no longer be necessary to use `-I` with any output mode except HTML, because HTML needs to find the source files.

EDIT: `-summary-only` does print point kinds.

aantron:
See the messages of the ~~last three~~ commits. ~~The other commits are from #106 and #107. I'd still like to wait until the end of the week for any objections to those, though I don't think there will be any.~~

cc @lindig

Resolves #102

@lindig: I have an urge to refactor a bunch of things in the instrumenter and reporter, including dealing with type definitions as you proposed in #100. If that would interfere with your work, I will wait – so let me know :)

Note: I tried several combinations in `InstrumentPpx.faster` to try to get the test `.reference` file diffs minimized, but `Pprintast` seems to have a mind of its own about inserting arbitrary whitespace far away from the affected expressions.

### DIFF
--- a/src/library/common.ml
+++ b/src/library/common.ml
@@ -51,22 +51,13 @@ exception Unsupported_version of string
 
 exception Modified_file of string
 
-let cmp_file_of_ml_file filename =
-  (if Filename.check_suffix filename ".ml" then
-    Filename.chop_suffix filename ".ml"
-  else
-    filename)
-  ^ ".cmp"
-
 let magic_number_rtd = "BISECT-RTD"
 
-let magic_number_pts = "BISECT-PTS"
-
 let supported_versions = [
-  1, 0
+  2, 0
 ]
 
-let format_version = (1, 0)
+let format_version = (2, 0)
 
 let write_channel channel magic write_digest x =
   output_string channel magic;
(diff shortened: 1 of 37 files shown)


aantron:
Ok, changing `read_points` and `read_runtime_data` breaks ocveralls, so I renamed the new functions `read_points'` and `read_runtime_data'`, and simulated the old functions on top of those. ocveralls now [works](https://coveralls.io/builds/6611882). However,

- Simulated `read_points` can only be called after simulated `read_runtime_data`, because `read_runtime_data` puts point definitions (former `.cmp`s) into a hash table keyed by source file name. This happens to be true in ocveralls.
- `read_runtime_data` sees file names before `-I` is applied, but `read_points` sees them with `-I` applied. To make the connection, I key using only the basename. This requires basenames to be unique within the project, which is usually true for OCaml.

Clearly, this can only be a temporary solution to pass tests. We will need to work with @sagotch to update ocveralls. It might also be helpful to hide all this stuff behind an even more abstract interface (@sagotch? @lindig?).

sagotch:
I did not read you patch yet so I do not know what is broken, but I will.

I will probably tell you more about this this week end.

lindig:
I haven't taken a closer look yet but I like the direction you are taking. In any case, it doesn't interfere with our use of Bisect PPX, so please go ahead.

Issue #112, opened 2016-08-12
build failure on OCaml 4.04


gasche:
opam-builder reports that `bisect_ppx` needs to be updated to work on OCaml 4.04: http://opam.ocamlpro.com/builder/html/bisect_ppx/bisect_ppx.1.1.0/7b6c8c5404fcbd94d5f63514ad51d9da

```
+ ocamlfind ocamlopt -c -package ppx_tools -w A-4 -warn-error A-3 -I src/syntax -I doc -I src -I src/report -I src/ocamlbuild -I src/library -o src/syntax/instrumentPpx.cmx src/syntax/instrumentPpx.ml
File "src/syntax/instrumentPpx.ml", line 152, characters 4-2061:
Warning 8: this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
Ppat_open (_, _)
File "src/syntax/instrumentPpx.ml", line 1:
Error: Some fatal warnings were triggered (1 occurrences)
Command exited with code 2.
```

rleonid:
@gasche Can you please provide some guidance on this new pattern `Ppat_open`. I haven't followed all of the forthcoming changes in 4.04.

rleonid:
Ah found it: https://github.com/ocaml/ocaml/pull/578.

aantron:
Fixed in #113.

PR #113, opened 2016-08-19
Extend marking to local open patterns

rleonid:
This PR is a WIP to allow bisect_ppx to work on 4.04. The testing for the new local open module feature (https://github.com/ocaml/ocaml/pull/578) isn't complete and unfortunately stalled because of [this](http://caml.inria.fr/mantis/view.php?id=7329) bug in the `-dsource` output.

Trying to fix issue #112


src/syntax/instrumentPpx.ml
```diff
@@ -208,6 +208,9 @@ let translate_pattern =
        it makes no sense to match a second time for effects such as
        exceptions. *)
     | Ppat_exception _ -> []
+    | Ppat_open (c, p') ->
+      translate mark p'
+      |> List.map (fun (marks, p'') -> marks, Pat.open_ ~loc ~attrs c p'')
```
aantron: The code change LGTM, but I would suggest that it be moved above the `Ppat_exception` case. Otherwise, it looks like the comment above applies to `Ppat_open` as well.

rleonid:
@aantron  Apologies for the delay, I was on vacation.

Updated the code per your requests and added tests. A couple of nits:
1. I'd suggest looking over the `.reference`, I did copy and paste the result and had a think about it. But this test is probably more important than the actual ppx-coverage rewriting instructions. I haven't started using these module open extensively (haven't switched to 4.04 yet), so a bit of a careful review would be helpful.
2. For some reason the new code doesn't get "covered" when looking at the result of `make coverage`... though it must be running.
3. I was a bit confused when my test code didn't compile (` assert c = M.foo`) and I couldn't figure out what was wrong. I don't know if you've had a similar experience but it felt like `oUnit` was being a bit obscure, but probably not worth doing anything.

aantron:
No problem, hope it was a nice vacation :)

Concerning 1 and 2, we currently don't call the pattern rewriter on patterns of `let` expressions (we only do it for `match`, `function`, and `try`). So, that is why the `Ppat_open` case is not being executed. The reference looks right, but it's also not testing this code as per the above.

Separately, we *should* probably consider calling the rewriter on `let`, since it allows or-patterns. I guess now we are lucky that we took the repeated-matching approach (nesting a second generated `match` expression) instead of promoting or-patterns to cases, since `let` doesn't allow cases. Might want to also look through the grammar for any other uses of production *pattern* we may have missed.

Re. 3, I tried introducing an error, and yes it's confusing. I don't remember the output being this bad before. Perhaps open an issue for making it better?

This change breaks compilation for OCaml < 4.04, so perhaps we need to rewrite the pattern matching in `translate_pattern` to use `ppx_tools` `ast_mapper_class`?

rleonid:
Now I understand. I couldn't help myself in extending the test case.

You make good points about the `let` differences.  I wouldn't worry too much about `3`, as long as the test-suite makes sense to you.

I'm not certain how to handle the `4.04` constraint, short of adding opam constraints on the latest bisect_ppx versions.`ppx_tools` isn't ready, and when I last looked at it for `bisect_ppx` it seemed like there were missing features, though I can't recall what they were at the moment.

aantron:
Would pinning `ppx_tools` work for now, for development? It looks like [this commit](https://github.com/alainfrisch/ppx_tools/commit/881a4826a4da482ceeca262950c5a8a0ca4e11d4) (and presumably its future back-port to 4.03) does what we want. IMO we should wait for `ppx_tools` to become ready, and put an `opam` constraint on that.

cc @alainfrisch

rleonid:
Thinking through the problem a bit more ... how would using `ast_mapper_class` help us with `translate_pattern` ? Specifically, wouldn't we still have to handle each pattern (`Ppat_*` ) case in our code?

I'm probably missing something.

aantron:
You're right, I misunderstood some `ppx_tools` code. We may have to add some kind of conditional compilation, or do parallel releases for >= 4.04 and < 4.04, unless `ppx_tools` adds something in `Ast_convenience` for dealing with this. `Ast_convenience` seems like the right way though, since we are probably not going to be the only ones with the problem.

rleonid:
> Ast_convenience seems like the right way though, since we are probably not going to be the only ones with the problem.

probably, but we'd have to mirror every pattern structure anyway, so why not use the `AST` directly?

>  or do parallel releases for >= 4.04 and < 4.04, 

Sure. Is there something I can add to this PR ?

aantron:
> probably, but we'd have to mirror every pattern structure anyway, so why not use the AST directly?

If we don't use the AST directly here, it is `ppx_tools` which would have to do parallel releases instead of us, which `ppx_tools` already does anyway, due to AST incompatibilities between 4.03 and 4.02. I hope I am understanding your concern, though :)

> Sure. Is there something I can add to this PR ?

I think it's probably best to open an issue on `ppx_tools` to see if it is likely to get `Ppat_open` compatibility functions implemented in `Ast_convenience`. I can also do a PR to `ppx_tools` later, in case you prefer that. Currently spending a lot of my open source attention on Lwt :/ Should probably put this PR on hold until after the `ppx_tools` discussion.

rleonid:
> put this PR on hold until after the ppx_tools discussion.

Fine with me, this is your wheelhouse.

aantron:
This will take a while to resolve on the `ppx_tools` side (https://github.com/alainfrisch/ppx_tools/pull/54#issuecomment-257241750). So, I think we should resolve the compatibility problem temporarily within Bisect_ppx, then switch to `ppx_tools`' solution once it becomes available.

I suppose you could adapt/inline the proposal in https://github.com/alainfrisch/ppx_tools/pull/54/commits/1a192b067672451572d7fcf005c30b990c048f05 and https://github.com/alainfrisch/ppx_tools/pull/55/commits/367ca8bab139008e4b836aa7de35cd98968d61df, or solve it any other way (e.g. temporarily depending on `cppo`).

I added testing on `4.04` to Travis (https://travis-ci.org/aantron/bisect_ppx/builds/172287998), so you should have that if you rebase.

rleonid:
@aantron Sorry for the delay on this one. I ended up using `cppo` to support multiple versions, and some hacking in the test suite but I think this would be a fine solution until the `ppx` libraries are ready.

tests/instrument/expr_match.ml
```diff
@@ -127,3 +127,4 @@ let f x =
 let f x =
   match x with
   | `Foo x | `Bar x -> x
+
```
aantron: Nit: is this adding a stray extra line?

tests/404_or_greater/expr_match.ml
```diff
@@ -0,0 +1,12 @@
+module M = struct
+  type t = Foo | Bar
+  type r = { i : int; t : t }
+end
+
+let f () =
+  let M.(s) = M.(Foo) in assert (s = M.Foo);
```
aantron: I believe the only fully useful tests right now would be `match` expressions, where at least one of the cases has an or-pattern, i.e. `M.(Foo | Bar)`, `M.{i = 3 | 4; _}`, etc.:

- We aren't running the pattern rewriter on `let` expressions at all (#115).
- If the rewriter generates only one case (because there are no or-patterns), we throw the case away and don't generate a nested `match` expression (https://github.com/aantron/bisect_ppx/blob/d587132aa537da793a669612d7e3f9c0d3c57fa9/src/syntax/instrumentPpx.ml#L255).

tests/test_helpers.ml
```diff
@@ -58,6 +58,24 @@ let _run_int command =
   | Unix.WEXITED v -> v
   | _ -> _command_failed command
 
+let _read_lines_from ~max_lines command =
```
aantron: I think it's possible to use [`Sys.ocaml_version`](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Sys.html#VALocaml_version) for this.

tests/404_or_greater/test_404_or_greater.ml
```diff
@@ -0,0 +1,4 @@
+open Test_helpers
+
+let tests =
+  compile_compare (fun () -> with_bisect () ^ " -w +A-32-4") "404_or_greater"
```
aantron: Nit: I think we should name the directory and test set as "`test_ppat_open`" or "`test_open_in_patterns`", or something like that.

- From the `4.04` feature list, it seems like the only new thing to deal with.
- If something else gets added in `4.05` or later, it won't go into `404_or_greater`.

tests/instrument/expr_match.ml
rleonid: eagle eyes

tests/404_or_greater/expr_match.ml
rleonid: True.
I'll add 
```OCaml
  let s = M.Bar in
  match s with
  | M.(Foo | Bar) -> assert true
  | _             -> assert false
```

 Do you want me to get rid off all the other stuff? Though we don't rewrite the pattern I can make a hand-wavy argument that it is nice to have since one could still write that type of code.

rleonid:
I'll squash past last comment about which tests to keep.
Regarding patching opam, might be easiest to do in PR to opam for a new version? I'd be glad to do it, but are there other things besides tagging that we follow to release a bisect_ppx ?

aantron:
Nope, pretty much just write changelog and tag (I wrote [`RELEASE.md`](https://github.com/aantron/bisect_ppx/blob/master/doc/RELEASING.md) to remind myself :p ).

Since this release would include some slightly messy feature removals by me, would you like to split it up this way: I'll take care of the release notes/tag, and you can do the PR to OPAM, with both the new release and the updates to older `opam` files, as you propose?

On a tangent, what do you think about stopping to update the `CHANGES` file? Just put a note in there, saying to see the GitHub releaes page?

tests/404_or_greater/expr_match.ml
aantron: I'd definitely convert all the syntax variants (`M.{}`, `M.[]`, `M.[||]`) to `match`, so that we are actually testing those, and insert an or-pattern in more of them. I'd personally delay adding `let` cases until #155... but they don't really hurt, besides making the future diff for #155 slightly less clear. But nobody is going to read that diff anyway after it's in `master`, so up to you :)

rleonid:
Squashed.

> Since this release would include some slightly messy feature removals by me, would you like to split it up this way: I'll take care of the release notes/tag, and you can do the PR to OPAM, with both the new release and the updates to older opam files, as you propose?

Sure, I'll look out for the tag.

> CHANGES.

I'm not too wedded to the file. But there is something nice to having the actual log be part of the source as opposed to being a `github` object.

aantron:
Thanks.

Tag is pushed.

And reasonable – let's keep `CHANGES`, then.

PR #117, opened 2017-04-13
Port to ocaml migrate parsetree, jbuilder

rgrinberg:
First of all, this gets rid of cppo!

A `ppx_bisect` module been added that will only register bisect with ocaml
migrate parsetree.

I've tried to maintain backwards compatibility by making the bisect_ppx be a
regular ppx mapper and adding `--as-ppx`.

The tests are still failing, because I'm guessing that I'm handling the command
line args incorrectly. Would be nice if someone could take a look and give some
pointers.

cc'ing the ppx wise men: @Drup, @let-def

TODO:

- [x] Fix installation. I'm not entirely sure what to do here, so I will need the advice of @rleonid and @aantron


opam/opam
```diff
@@ -22,7 +22,7 @@ depends: [
   "ocamlfind" {build}
   "ppx_tools" {build & >= "4.02.3"}
```
Drup: You want to use ppx_tools_versioned too.

src/META
```diff
@@ -1,7 +1,7 @@
 version="1.2.0"
 description="Code coverage for OCaml"
 requires="bisect_ppx.runtime"
-ppx="./bisect_ppx"
+ppx="./bisect_ppx --as-ppx"
```
Drup: The META should contain various additional things to play nicely with the omp driver. There are exemples in omp's repository.

src/syntax/ppx_bisect.ml
```diff
@@ -0,0 +1,13 @@
...
+
+let () =
+  Driver.register ~name:"ppx_bisect" ~args:InstrumentArgs.switches
+    Versions.ocaml_404 (fun _config _cookies ->
+        (* let anon s = raise (invalid_arg ("nothing anonymous: " ^ s)) in *)
+        let anon _s = () in
+        let usage = Printf.sprintf "Usage: bisect_ppx <options>" in
+        let arga = Array.of_list ("" :: (Array.to_list Sys.argv)) in
+        Arg.parse_argv arga InstrumentArgs.switches anon usage;
+        Ast_mapper_class.to_mapper (new InstrumentPpx.instrumenter)
```
Drup: I will let @let-def comment, but I doubt that will fly with omp's driver. :p

src/syntax/ppx_bisect.ml
```diff
@@ -0,0 +1,13 @@
+open Migrate_parsetree
+open Ppx_tools_404
+
+let () =
+  Driver.register ~name:"ppx_bisect" ~args:InstrumentArgs.switches
+    Versions.ocaml_404 (fun _config _cookies ->
+        (* let anon s = raise (invalid_arg ("nothing anonymous: " ^ s)) in *)
+        let anon _s = () in
```
rleonid: Is there a reason for this change in `anon`'s behavior?

rgrinberg: Simply because I wasn't even able to run `./bisect_ppx --as-ppx` at all without removing this exception. I'll fix it up once I understand how to handle args correctly with omp

rgrinberg: Looks like omp disallows anonymous arguments anyway btw:

https://github.com/let-def/ocaml-migrate-parsetree/blob/master/MANUAL.md#using-arguments-in-meta-ppxopt

src/syntax/ppx_bisect.ml
let-def: Yes, the purpose of exposing Arg.specs to the driver is that the rewriter doesn't deal with arg parsing by itself (so that multiple rewriters are more likely to compose).

So you shouldn't do `Arg.parse_argv`, when the rewriting function is called Arg.specs have already been executed.

src/travis_ci_test.sh
```diff
@@ -58,7 +58,7 @@ git --version
 echo
 echo "Install dependencies"
 echo
-opam install -y ocamlfind ocamlbuild ppx_tools cppo
+opam install -y ocamlfind ocamlbuild ocaml-migrate-parsetree
```
let-def: Maybe `ppx_tools_versioned` too?

aantron:
I re-ran the 4.02 Linux build, which seemed to have failed with a spurious failure to build `jbuilder`. It now [fails](https://travis-ci.org/aantron/bisect_ppx/jobs/221913903#L353) with

```
File "src/syntax/ppx_bisect.ml", line 1, characters 5-22:
Error: Unbound module Migrate_parsetree
```

I ran twice to confirm. I'm not sure why that would happen on Linux and not macOS...

The 4.03 failure on macOS can be ignored; that's due to me trying to be clever in the CI script, and trying to use Homebrew's system OCaml to save time. They have since moved to 4.04. I will fix that later, separately from this PR.

rgrinberg:
@aantron I pushed a commit that should now make it build for you.

aantron:
Thanks for the rebase. @rgrinberg @bradlangel any update on the status of this? (no rush!)

rgrinberg:
@aantron the META seems fine but tweaking that build system as hell.

I'm thinking you guys should just get rid of the self instrumenting and switch to jbuilder :cry:

aantron:
If that makes it easier, sure. We can figure out how to add it back later. It's not that important to our users.

rgrinberg:
It makes it much easier for me as I don't have to figure out the current build system. But if you know how to fix what we have now, perhaps that is easier for you.

I'm definitely casting my vote towards jbuilder though.

rgrinberg:
It makes it much easier for me as I don't have to figure out the current build system. But if you know how to fix what we have now, perhaps that is easier for you.

I'm definitely casting my vote towards jbuilder though.

aantron:
Then port to jbuilder (please :)). Don't let the self-instrumentation stand in the way. We will figure out later how to add it back, or just forget it. There are many different ways to add it back.

rgrinberg:
OK, the library is ported to jbuilder except for the tests.

The package naming is a bit bizarre. bisect_ppx is the runtime package while ppx_bisect is the preprocessor. Really strange...

IMO you should simply take over the bisect opam package

aantron:
Thanks! Do you plan to port the tests as well, or leaving this for us? :)

Taking over the `bisect` package would be nice, but it is still about even for downloads with `bisect_ppx` [[1](http://opam.ocaml.org/packages/bisect/)] [[2](http://opam.ocaml.org/packages/bisect_ppx/)], and I suspect some of the users are using it with Camlp4. On the other hand, since `bisect` is not being updated in OPAM, those users could just constrain themselves to a version. It does seem like it would be unnecessarily disruptive, though.

rgrinberg:
I got the tests to build. At this point my progress there is slowing down as I haven't really understood the structure behind the tests. Would definitely appreciate some hints or you taking over.

As for naming:

The way I see it, every library should maintain a 1 to 1 findlib opam mapping. To me this suggests one of two alternatives:

```
ppx_bisect # bisect preprocessor
bisect # runtime
bisect-ocamlbuild
bisect_ppx # compatibility package
```

Or:

```
ppx_bisect # preprocessor
bisect-runtime # runtime
bisect-ocamlbuild
bisect_ppx # a compatibility package
```

The presence of that camlp4 package seems to force our hand into option 2. Is that ok?

rgrinberg:
A small point about porting the tests:

It seems like the tests are preprocessing with bisect to compare the output using the compiler. But with jbuilder this is much easier (as all things). Simply invoked `_build/default/.ppx/ppx_bisect/ppx.exe`. Here are the options:

```
ppx.exe [options] [<files>]
  --as-ppx                  Act as a -ppx rewriter
  --as-pp                   Shorthand for: --dump-ast --embed-errors
  --dump-ast                Output a binary AST instead of source code
  -o FILE                   Output to this file instead of the standard output
  --intf FILE               Treat FILE as a .mli file
  --impl FILE               Treat FILE as a .ml file
  --cookie NAME=EXPR        Set the cookie NAME to EXPR
  --embed-errors            Embed error reported by rewriters into the AST
  -exclude <pattern>        Exclude functions matching pattern
  -exclude-file <filename>  Exclude functions listed in given file
  -mode {safe|fast|faster}
                            Ignored
  -runtime <module          name>  Set runtime module name; used for testing
  -simple-cases             Do not generate separate points for clauses of or-patterns
  -inexhaustive-matching    Generate inexhaustive match expressions in cases; used for testing
  -help                     Display this list of options
  --help                    Display this list of options
```

rleonid:
@rgrinberg What does this

> bisect_ppx # compatibility package

package do? Sorry that I'm not following along.

rgrinberg:
The compatibility package would depend on all other bisect packages and would only install a META file as a compatibility shim. So it would include the `runtime`, `fast`, `ocamlbuild`, and  `plugin` findlib subpackages which would just `requires` on the real packages listed above.

aantron:
@rgrinberg Each test takes an `.ml` file, preprocesses it with Bisect, and then dumps the resulting AST using `Printast` (or `Pprintast`, I forgot which), which formats it as OCaml source.

We then compare the dumped AST to a reference dump (`.ml.reference`). However, the exact formatting of `Printast`/`Pprintast` differs between versions, so what we actually do for the reference is have the compiler parse it, then dump it again. That way, we are comparing preprocessed and reference dumps done by the same version of the compiler (the installed one). Does that help?

I can take over this, but a bit later, as I have a backlog :/

aantron:
Ok, working on it from now.

aantron:
I spent several hours trying to fix the tests, then decided to do what any good developer would do: disable them completely. Given that Bisect_ppx doesn't see a high rate of development, I think we can survive a few days or weeks without a test suite.

We should just merge this PR if it works with existing projects. We (I?) can sort out follow-on problems later. I want to avoid any more bit rot of this PR, and move forward.

I pinned the `bisect_ppx` in this branch and installed it. It doesn't work with the Lwt `working-coverage` branch. Examining the installed `META` file, I think that the old-style PPX executable is not running:

```
version = "dev"
description = ""
requires = ""
archive(byte) = "bisect.cma"
archive(native) = "bisect.cmxa"
plugin(byte) = "bisect.cma"
plugin(native) = "bisect.cmxs"
```

And all the old-time packages are missing. Here is the `META` from Bisect_ppx 1.2.0:

```
version="1.2.0"
description="Code coverage for OCaml"
requires="bisect_ppx.runtime"
ppx="./bisect_ppx"

package "runtime" (
  version="1.2.0"
  description="Code coverage for OCaml"
  archive(byte)="bisect.cma"
  archive(native)="bisect.cmxa"
)

package "fast" (
  version="1.2.0"
  description="Code coverage for OCaml (deprecated package)"
  requires="bisect_ppx"
)

package "ocamlbuild" (
  version="1.2.0"
  description="Bisect_ppx Ocamlbuild plugin"
  archive(byte)="bisect_ppx_plugin.cma"
  archive(native)="bisect_ppx_plugin.cmxa"
)

package "plugin" (
  version="1.2.0"
  description="Bisect_ppx Ocamlbuild plugin (deprecated package)"
  requires="bisect_ppx.ocamlbuild"
)
```

@rgrinberg, can you address this? Then, we will merge the PR.

`bisect-ppx-report` appears to be getting installed correctly.

aantron:
For the sake of completeness, "does not work" means no `.out` files were generated, presumably because the code was not actually instrumented.

rleonid:
Without porting all of the tests, Is there a way in `jbuilder` to exec arbitrary code or processes? I'm wondering if we can just leave the test suite as is and when testing have `jbuilder` call `make test`.

aantron:
I believe the current problem with the tests is that they expect Bisect to be installed in a `_findlib` directory, so they won't work until Bisect installation is fixed. Otherwise, I think they already behave as they did before, except that there are a few more lines of code to remove, to fully get rid of self-instrumentation.

aantron:
Just an update: this PR now has Jbuilder-based backwards-compatible Bisect, and I believe it works with all pre-existing non-Jbuilder Bisect users. Bisect also "works" with Jbuilder, subject to the wrinkle that you kind of have to hardcode it into user projects until https://github.com/janestreet/jbuilder/issues/57 is addressed somehow.

The `ppx_bisect` package that this PR had introduced is now just the root `bisect_ppx`; I'm using Jbuilder's "deprecated-ppx-method" mechanism (in the generated `META`) so that we don't need to tell users, at this point, to use one of two different package names depending on how their PPX is run. See the [example of this that is checked into Lwt](https://github.com/ocsigen/lwt/blob/cc98e5f09383a6a465d3aa9620ddf2d46eb16075/META.lwt.template#L26-L52).

Because the Bisect in this PR is no longer using Ocamlbuild, I went against best practices, and made `ocamlbuild` a depopt of `bisect_ppx` for `bisect_ppx.ocamlbuild`. This is supposed to be a temporary state, until some future release where we factor out `bisect_ppx-ocamlbuild` into its own OPAM package, as was originally done in this PR. We should also get rid of the `bisect_ppx.fast` and `bisect_ppx.plugin` deprecated alias packages at that time. In general, we can mutate the Bisect_ppx packaging in any direction, including introducing `ppx_bisect`, but I want to avoid mixing such strides into this already-involved PR.

In summary, from the point of view of users, I believe this PR is done.

What's left is some more thorough packaging testing. Also, I now believe it shouldn't be hard to port the test suite, so I will try again to do that too. Then, I will resolve the merge conflicts, and do a squash-merge. The main author will be @rgrinberg, and I will note that @bradlangel and myself did some additional work in the message. Please let me know if you have any objections.

Also, @edwintorok, I'm sorry that this has held up the release. I'm working too slow :p

rgrinberg:
Thanks for all the work Anton. This looks good. Looking forward to reshuffling the packaging of bisect in the future.

aantron:
@rginberg I fixed the majority of the tests, however note the last commit about the ppxopt tests:

```
Passing ppxopts used to be done with ocamlfind by

  -package bisect_ppx -ppxopt bisect_ppx,your_opts

with the bisect_ppx.deprecated-ppx-method scheme, it apparently has to
be, at least for now,

  -package bisect_ppx -ppxopt bisect_ppx.deprecated-ppx-method,your_opts

This makes the tests pass, but isn't really acceptable for release.

Tests with ppx_deriving and ppx_blob still remain.
```

This is in reference to [Jbuilder's `deprecated-ppx-method` scheme](https://github.com/ocsigen/lwt/blob/cc98e5f09383a6a465d3aa9620ddf2d46eb16075/META.lwt.template#L26-L52) referenced above.

Do you know if there is some workaround for this? I can open an issue on Jbuilder if not. This will likely block the Bisect_ppx release, unless I restore the `ppx_bisect` package, which seems like a pretty excessive workaround for what may be a minor oversight.

rgrinberg:
OK  I see. No I'm not aware of a workaround unfortunately. Might I suggest that we reconsider making breaking changes here perhaps and getting the packaging right out of the box?

There aren't that many disadvantages to breaking stuff IMO. ppx_bisect isn't a dependency of any package and hence isn't going to cause downstream breakage. So in the worst case, it will cause an hour's worth of inconvenience for users (which can be very easily mitigated with a good write up in the readme) or just make them stick to an older version of bisect for now.

I understand and appreciate your commitment to backwards compat, but I don't really think this situation is analogous to Lwt.

aantron:
Yes, we can fix the packaging that way, and I'm not fully against it. I realize we probably won't affect anyone, except for the perpetual minor confusion in the future. But I would prefer not to *have* to do it – the only reason we would be forced to is because Jbuilder/omp backwards compatibility isn't well-enough thought out (I don't want to overstate this; it's only on this minor oversight).

As of now, Bisect doesn't seem to fit into the Jbuilder model anyway (or vice versa). Since we have to press for fixing that, it seems better to also press for this to be resolved somehow. This is especially given that there seems to have been an effort with `deprecated-ppx-method` in the first place, it's just incomplete. I'll give it some thought later. Maybe a smarter `META` file is the right way (inverting the packages, or using three?).

Actually, for most Jbuilder users, there is no point in releasing this PR for now anyway, given the lack of a good Bisect/Jbuilder usability story. So, I was aiming at the non-Jbuilder users, hence the extra emphasis on backwards compatibility. Ideally, we are able to only swap out the build system internally in Bisect, add omp support, and be ready for Jbuilder.

I mainly wanted to release this so I could hardcode Bisect into development Lwt, since it uses Jbuilder now. It's also nice not to let this PR keep bit rotting... :)

Jbuilder is causing pain. I know it has many advantages, but the blind spots are causing non-trivial maintenance headaches.

aantron:
The packaging snag is presumably worked around now, and we can keep the single `bisect_ppx` package name for the time being. See https://github.com/rgrinberg/bisect_ppx/blob/70242ac480b61cf6e29958d7a64795c622f728a2/META.bisect_ppx#L12-L35. Please let me know if anything there looks wrong.

aantron:
😓  thanks!

Bisect_ppx is Jbuilder-ized.

aantron:
...and ocaml-migrate-parsetree-ified.

PR #118, opened 2017-05-14
fix instrumentation of refutation clauses `p -> .`

gasche:
[refutation clauses](https://caml.inria.fr/pub/docs/manual-ocaml-4.04/extn.html#sec241) `| p -> .` must not get instrumented, as
instrumentation would generate code of the form

     | p -> <instrumentation>; .

that makes the type-checker fail with an error as it does not
recognize the refutation clause anymore:

    File "test.ml", line 3, characters 2-4:
    Warning 11: this match case is unused.
    File "test.ml", line 3, characters 8-9:
    Error: Unreachable expression was reached

The type-checker could in principle be extended to recognize
right-hand-sides of the form (e; .) as statically unreachable, but
this fails at least with OCaml 4.03 and 4.04.

Note that this patch could be trivially extended to automatically
consider clauses of the form

    | p -> assert false

as unreachable, which I would personally find desirable as they are
(in my code) always a clear intent of intentional unreachability
(so I would never write a test to cover them, and would have to
manually add an excluding comment). The current instrumenter does not
exclude these, so I did not implement it.


gasche:
Remarks:

- this is a bisect-breaks-the-compilation-of-my-project bug
- I could be conviced to rebase on top of #117 if you merged it

aantron:
I think fixing this in Bisect is indeed the right thing to do.

I actually had a TODO item to not mark `assert false`, just never got around to it. If it's not too much trouble for you to add that, it would be welcome.

I'll give this a proper review in a day or two. Thanks for the PR!

gasche:
What is very easy to do within this patch is to consider the whole clause `| p -> assert false` as unreachable. Did you had in mind a more general behavior where any `assert false` would be considered unreachable? (I am not convinced the more general behavior is a good idea: if people write stuff like `(<expr>; assert false)`, it means that either they have dead code and they should be told, or they consider that this `assert false` may be reachable.)

aantron:
Just handling the complete clauses would be great already, I think that's a pretty common scenario that we can easily take care of. I would benefit from it in Lwt, for instance (and it's how it ended up on my TODO list in the first place).

I guess the 4.02 builds are breaking because refutation clauses require >= 4.03. Do you want me to look at the test suite later, to execute those tests conditionally? You should be able to put something together using [`Test_helpers.ocamlc_version`](https://github.com/aantron/bisect_ppx/blob/master/tests/test_helpers.mli#L42). I think we already have some tests that run conditionally, I just don't remember off the top of my head which ones (some PPX tests, IIRC).

aantron:
And please ignore the 4.03 failure, that's totally my fault, and I'll take care of it later (soon!). Sorry about that.

gasche:
I added the logic for `| p -> assert false` in a separate commit. This changes the behavior of the test `ppat_open/expr_match.ml` in a way that seems reasonable.

I hadn't thought of the testsuite under 4.02, I'll go fix it.

gasche:
Rebased on top of #121 for version-conditional tests.

aantron:
@gasche Thanks for the bug report/pull requests. I especially appreciate the added tests and detailed commit messages :) I'll take care of merging with the `ocaml-migrate-parsetree` PR.

I've added a commit to `master`, after these PRs, that should fix macOS testing on 4.03.

Now, on to the other PR...

PR #119, opened 2017-05-15
Minor changes

gasche:
This PR does not change the tool behavior, but it adds a testsuite for IGNORE/VISIT patterns and their interaction with clause marking, and reorder the instrumented code to make the output readable (I would checked that I got the expected VISIT behavior, which requires looking at the generated code.) Feel free to kill this PR under the "if it ain't broke, don't fix it" rule.

### DIFF
--- a/src/syntax/instrumentPpx.cppo.ml
+++ b/src/syntax/instrumentPpx.cppo.ml
@@ -330,14 +330,20 @@ let faster file =
   in
   let make = apply_nolabs (lid "Array.make") [intconst nb; intconst 0] in
   let marks =
-    List.fold_left
-      (fun acc (idx, nb) ->
-        let mark =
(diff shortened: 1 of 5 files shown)


aantron:
> Feel free to kill this PR under the "if it ain't broke, don't fix it" rule.

On the contrary, this is a welcome change. Thank you :)

Some notes:

- The test suite is failing with

        File "test_main.ml", line 31, characters 2-29:
        Error: Unbound module Test_exclude_comments

- `master` now has a fixed macOS 4.03 build. While not necessary, it may be nice to rebase over that.

I know your time is precious (thanks for all your other work!), so please let me know if you'd like me to take over this and make the minor changes needed to fix it up. I'll check with you for proper blame/authorship before merging anything.

gasche:
Hm, I forgot to `git add` a file ^^ This should be fixed now, thanks.

aantron:
Thanks again.

PR #121, opened 2017-05-15
Conditional tests

gasche:
The existing machinery only allows to choose a minimum version for a whole directory, not per-file.


tests/test_helpers.mli
```diff
@@ -42,8 +42,11 @@ val have_package : string -> bool
 val ocamlc_version : unit -> int * int * int option
 (** Return the [ocamlc] version. *)
 
+val ocamlc_403_or_more : unit -> bool
+(** [ocamlc] version is 4.03.0 ir greater. *)
```
aantron: This `ir` outlived the fix of the other one (thanks) :)

tests/test_helpers.ml
```diff
@@ -174,6 +174,9 @@ let ocamlc_version () =
     (fun major minor _periods patchlevel ->
         major, minor, try Some (int_of_string patchlevel) with _ -> None)
 
+let ocamlc_403_or_more () =
+  ocamlc_version () >= (4,4,None)
```
aantron: I guess the second `4` should be a `3`?

gasche: Fxed, thanks.

PR #123, opened 2017-05-20
Fixes unary operator expected issue

bradlangel:
When running `make install` the following error occurred:
`/bin/sh: line 0: [: =: unary operator expected`
In makefile we need to put our reference to DEV_INSTALL in double
quotes.


aantron:
Thanks!

Issue #122, opened 2017-05-16
dumping coverage data based on a trigger


edwintorok:
In some cases it might be useful to dump coverage data sooner than `at_exit` (for C applications you could call `__gcov_flush` to do that).
For example: long running daemon that needs to stay alive to run more testcases.

Signal handling might be an option (optionally based on an env var), if both bisect_ppx and the actual program take care of calling the old signal handler in the newly installed one.
Another one would be listening on a socket/fifo but that requires the Thread module.

What are your thoughts? Would a PR that adds a signal handler be good?

aantron:
That would be pretty useful indeed.

My guess would be that Bisect can't make decisions about threads or signal handlers inside its own runtime. That's too likely to interfere with applications.

It should probably expose some variant of [[this code]](https://github.com/aantron/bisect_ppx/blob/master/src/library/runtime.ml#L81-L83) somehow, and document it. The application/tester can then call that however it sees fit.

What do you think?

edwintorok:
Yes, it would be better to expose a function in the mli. Although using it would require some sort of conditional compilation, so that it won't fail if `Bisect.Runtime.write_dump` is not available when not compiling with coverage.

aantron:
That's a good point. I forgot about this complication. Not sure what the best thing to do here is, but setting an environment variable to a signal number or fd now sounds much more reasonable.

There is some talk of changing the license. Maybe that will allow always linking Bisect, but provide a no-op runtime when coverage is disabled. I think that should work even in the presence of dependencies – `Bisect.Runtime.write_dump` should only need to be called by top-level applications and test runners. In the case of applications/daemons that directly call `write_dump`, it would make Bisect no longer only a `{test}` dependency, however.

aantron:
IIRC this was resolved by #129, so closing it. Please let me know if that's inaccurate.

PR #129, opened 2017-06-02
Expose functions in runtime.mli to dump counters

edwintorok:
Beginning of support for https://github.com/aantron/bisect_ppx/issues/122
Not sure whether `dump_counters_exn` should take a channel, or just take a `unit` and return a `string` with the filename (and keep using `file_channel` internally). I'd slightly favour giving more control to the user of bisect_ppx, i.e. the channel param. Thoughts?

### DIFF
--- a/src/library/runtime.ml
+++ b/src/library/runtime.ml
@@ -74,13 +74,23 @@ let file_channel () =
   let channel_opt = ic_opt_loop (next_name ()) in
   channel_opt
 
+let dump_counters_exn channel =
+  let content = Hashtbl.fold (fun k v acc -> (k, v) :: acc) table [] in
+  Common.write_runtime_data channel content
+
+let reset_counters () =
+  Hashtbl.iter (fun _ (marks, _) ->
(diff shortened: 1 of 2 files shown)


aantron:
This looks good. I agree about using a `channel` rather than filename. In fact, I'd be concerned about stdlib channels not being general enough, but we don't need to spend time addressing that now IMO. @rleonid, thoughts?

@edwintorok I restarted the stuck build. Let's see what happens.

rleonid:
Looks good to me

aantron:
Thanks @edwintorok, @rleonid

edwintorok:
Thanks a lot for merging. Are you planning to make a release containing this change?
I'd like to add a version constraint in my code when using the new API.

aantron:
Yes. I'm going to work in the coming day or two on getting the jbuilder/omp PR (#117) finished/reviewed and merged. Then, tie some minor loose ends and we should release. I would say that should happen within a week or two.

PR #137, opened 2017-06-16
[report] Provide useful information in case of `input_value` error.

ejgallego:
Thanks for bisect_ppx. In some cases when my build is interrupted, I get some corrupted files but, I cannot determine which one is the bad one as the exception lacks information. This patch adds the bad filename to error printing.


aantron:
Thanks!

PR #139, opened 2017-06-17
Uses source filenames for the html-report urls

### LINKED ISSUE #124: Use source filenames instead of numbers in URLs
The filenames of the Bisect_ppx HTML report files currently look like

    http://aantron.github.io/bisect_ppx/coverage/file0014.html

I think this is designed to survive renaming, but I think it would make more sense if Bisect_ppx produced

    http://aantron.github.io/bisect_ppx/coverage/instrumentPpx.ml.html

or something along those lines.

I think for this, it is sufficient to edit [this line](https://github.com/aantron/bisect_ppx/blob/18e048db4bda3b3128900c22bb84483f4d1d4f48/src/report/reportHTML.ml#L711). I think `in_file` there is just the string path to the source file, so you need to call `Filename.split_filename` or something on it. Then, generate a Bisect_ppx report and make sure the files are named nicely, and the links from the report main page still work.

I don't think this will break any users.

bradlangel:
Summary:
Modifies the .html files output by bisect-ppx-report to use the source
filename rather than an arbitrary number. For example:
`file0001.html` becomes `source.ml.html`.
- fixes #124

### DIFF
--- a/src/report/report_html.ml
+++ b/src/report/report_html.ml
@@ -696,8 +696,7 @@ let output_html
 let output verbose dir tab_size title resolver data points =
   let files = Hashtbl.fold
       (fun in_file visited acc ->
-        let l = List.length acc in
-        let basename = Printf.sprintf "file%04d" l in
+        let basename = Filename.basename in_file in
         let out_file = (Filename.concat dir basename) ^ ".html" in
         let maybe_stats =
           output_html verbose tab_size title in_file out_file resolver
--- a/src/report/report_html.ml
+++ b/src/report/report_html.ml
@@ -694,7 +694,8 @@ let output_html
     Some stats
 
 let output verbose dir tab_size title resolver data points =
-  let files = Hashtbl.fold
+  let files =
+    Hashtbl.fold
       (fun in_file visited acc ->
         let basename = Filename.basename in_file in
         let out_file = (Filename.concat dir basename) ^ ".html" in
--- a/test/src/test_report.ml
+++ b/test/src/test_report.ml
@@ -26,7 +26,7 @@ let tests = "report" >::: [
 
   test "html" (fun () ->
     report "-html html_dir";
-    run "grep -v 'id=\"footer\"' html_dir/file0000.html > output";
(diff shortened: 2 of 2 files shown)


aantron:
We may have to switch to full relative paths in the future, but I think this is good for now. IMO it's a definite improvement over file numbers.

I split out the whitespace-only changes (except for the indentation of one line, which I dropped) into their own commit. It's a bit of maintainer tedium, so I hope that's okay. It helps keep the history more legible, as someone looking in blame can see, at a glance, that the whitespace-only commit affected no logic, and click past it. You can review the diff by running

```
git fetch origin
git diff source-filenames origin/source-filenames
```

and if you want to restore the history as it was,

```
git push -f
```

from `source-filenames`.

bradlangel:
>We may have to switch to full relative paths in the future, but I think this is good for now. IMO it's a definite improvement over file numbers.

I didn't think about this but we may just want to do it now. This change could potentially cause an issue in people's coverage if their html files are being overwritten. I think the only thing blocking us from doing this is that we get an exception that the directory doesn't exist when it tries to the write the file. I'll look into it some more this week.

>It's a bit of maintainer tedium, so I hope that's okay

Yeah, it is definitely okay. I'm sorry you had to modify the commit as a result of my commit sloppiness. Thanks for doing that!

aantron:
Sure, and no problem. I do it all the time to my own commits (it's easy to write a bunch of stuff, then figure out how to split it). Let me know when it's ready for another look :)

brendanlong:
Was this basically merged? Should this pull request be closed now or is there still something that needs to be merged here?

aantron:
From my point of view, it's ready for merge. If @bradlangel doesn't mind, I'll rebase the PR to fix the merge conflict, and merge.

Immediately after merging, we probably need to switch Bisect_ppx to use full relative paths for the HTML files, instead of just filenames. Currently, with Jbuilder, it is pretty common for the same filename to appear in multiple different directories of the same project.

aantron:
This has been released in Bisect_ppx 1.3.2. Thanks @bradlangel, and @brendanlong for the prod.

Issue #143, opened 2017-06-29
Factor out bisect_ppx-ocamlbuild as a separate OPAM package


aantron:
This is to eliminate the [depopt](https://github.com/aantron/bisect_ppx/blob/07a994f6bbe8724ad62e77b75f2c326810da33ea/bisect_ppx.opam#L33) :scream: we temporarily got as part of #117, and is a follow-on to that PR. We should release once with the depopt, but with the new package, and warn people using the Ocamlbuild plugin to switch to it.

PR #146, opened 2017-07-03
Begin factoring out bisect_ppx-ocamlbuild

aantron:
This creates a new OPAM package `bisect_ppx-ocamlbuild`, which contains a single new Findlib package with the same name. That Findlib package is currently just an alias for `bisect_ppx.ocamlbuild`, provided by existing OPAM package `bisect_ppx`. This is a transitionary state. In the future, the Ocamlbuild plugin will be moved completely into this new OPAM package.

At that point, we should be able to delete the custom `META.bisect_ppx-ocamlbuild`. We should also then submit a PR to OPAM to constrain the initial release of `bisect_ppx-ocamlbuild` to require at most the last version of Bisect_ppx that still included the plugin directly.

We should post an issue listing the expected breakage in Bisect_ppx 2.0.0, and put a link to it in the `bisect_ppx.opam` post-install message.

Part of #143.

cc @rgrinberg
cc @andrewray for FYI; we will have to do some similar things in Lwt. This isn't really critical to do in Bisect, it's more like practice for Lwt :)

The rest of this is a list of packages or repos that I found that use the old packaging of the plugin. I didn't cc the maintainers yet, because until we release this PR for the first time, there is no action for the maintainers to take. We will cc them upon release.

https://github.com/objmagic/synsyn (objmagic)
https://github.com/reactorlabs/sourir (O-, gasche)
https://github.com/edwintorok/xapi-work-queues (edwintorok)
https://github.com/briancaine/yas3cmd (briancaine)
https://github.com/xapi-project/xenopsd (edwintorok)
https://github.com/edwintorok/xapi-work-queues (edwintorok)

PS the GIF uses the old package name, but that's not a big deal :) We'll get a new GIF at some point anyway.


bisect_ppx-ocamlbuild.opam
```diff
@@ -0,0 +1,24 @@
+name: "bisect_ppx-ocamlbuild"
```
rgrinberg: Note that it's not necessary to add `name:`  if the name is already included in the filename.

aantron: I guess we won't be renaming the file to `opam`, so we can remove the `name:` fields.

aantron: @rgrinberg should `version` also be avoided? I guess it's always known when installing from OPAM, and when pinning it's inferred as `dev` anyway AFAIK.

rgrinberg: From what @Drup told me `version: dev` is required for opam to show the correct version when pinning. But I never noticed the difference myself.

Drup: @aantron No, pinning doesn't imply dev (why should it ? You can pin to a source package containing other versions).

aantron: Yes, but I thought those might contain `version:`, but the development ones don't need it. Anyway, I just tried it and it doesn't work. Not including `version:` results in version `~unknown` pinned, not `dev`. Thanks.

hcarty: Having `dev` as a version in a pinned opam file has strange effects like causing upper bounds to break. I'm not sure if the default `~unknown` has the same effect.

aantron: I guess breaking upper bounds is a problem in some cases. I usually want them to break though, as I'm trying to install a dev version "no matter what," so I assumed this was deliberate.

Drup: In my opinion, breaking upper bound for the dev version is precisely the intended effect.

hcarty: Sorry, by breaking an upper bound I mean that `dev` ends up violating a `< 4.0.0` version constraint for example, preventing pinning against a git repo.

hcarty: I also want to be able to pin against some development state and have it "just work" without worrying about version constraints. Unfortunately using `dev` doesn't seem to provide that.

aantron: Actually, that's right. I had to resort to something like this: https://github.com/ocsigen/lwt/blob/611ef1833c7aa0ce53221c7589b2673cc17d3063/lwt_ssl.opam#L21 which only "worked" because the constraint syntax was incorrect.

hcarty: I ran into this recently when testing against Lwt's latest master branch - I had a package installed that had an (arguably overly pessimistic) constraint of `lwt < 4.0.0` so the pin didn't work.  I had to modify the other package to remove the upper bound on the Lwt version constraint.

aantron: @hcarty out of curiosity, if the package is public, what is it?

hcarty: @aantron Not public unfortunately - it's an internal tool.   I can make an example project under my github account though which illustrates the issue if that would help.

aantron: It's okay, I already know it from having those subpackages :) I was just wondering if this usage is leaking into the open wild, in which case it may be good to make an announcement to stop it.

hcarty: I think it would be good to make such an announcement.  The error message is somewhat unclear (Why does pinning break my code?  What does it mean that dev doesn't match < 4.0.0?) and the fix is painful.  I have a package waiting to be released which I removed all of the upper version bounds from because it was too much of a pain to deal with all of the conflicts when pins are involved.

Should these `version: dev` lines be actively removed?  In my experience so far they cause more harm than good.  I can start something on discuss.ocaml.org if you're interested in chiming in.

aantron: I actually have a TODO item to write a brief "Depending" section in the README, and then announce that together with an explanation of how the Lwt breakage process works (with the major releases and warnings, etc.). Ideally, more projects would adopt that... but at a minimum, more people would subscribe to the announcements issue.

Does having `~unknown` as a version not have the problem of interacting with upper bounds? If so, I think you should start that thread... or if you can come up with any other good alternative.

aantron:
In the future Bisect_ppx 2.0.0 (at least three or so months away), `bisect_ppx.ocamlbuild` will no longer be a valid package name.

Instead, `bisect_ppx-ocamlbuild` should be used (dot replaced by hyphen). This is provided by a new, separate OPAM package with that same name `bisect_ppx-ocamlbuild`. In other words, the Bisect_ppx Ocamlbuild plugin is being factored out into its own OPAM package.

To ease the transition, both package names are available for now, since today's release of Bisect_ppx 1.3.0. I'm ccing maintainers of code I found using the deprecated, to-be-removed package name:

> https://github.com/objmagic/synsyn (@objmagic)
> https://github.com/reactorlabs/sourir (@O-, @gasche)
> https://github.com/edwintorok/xapi-work-queues (@edwintorok)
> https://github.com/briancaine/yas3cmd (@briancaine)
> https://github.com/xapi-project/xenopsd (@edwintorok)
> https://github.com/edwintorok/xapi-work-queues (@edwintorok)

edwintorok:
Thanks, the projects you mentioned with my name on it will probably get transitioned to jbuilder, so no worries.

aantron:
@edwintorok there is also nicer Jbuilder support now – just want to make sure you've seen it :) https://github.com/aantron/bisect_ppx/blob/master/doc/advanced.md#Jbuilder

gasche:
I did the name change, thanks for the notice.

briancaine:
Thanks for the heads up.

Issue #150, opened 2017-08-03
Cannot find opam package bisect_ppx-ocamlbuild


KasperJanssens:
The instructions in the Readme.md suggest to install bisect_ppx-ocamlbuild when working with myocamlbuild. However, I am not able to find this package in the opam repository (https://opam.ocaml.org/packages/). So I was wondering then where I should find it? Or is it not necessary anymore?

This is more of a question than a bug of course.

aantron:
@KasperJanssens It's not necessary *yet*. Sorry about that. The plan was to release `bisect_ppx` 1.3.0 and concurrently `bisect_ppx-ocamlbuild`, but life and other work is interfering, so there is a delay :/

For now, if you install `bisect_ppx` alone, it's enough to get the plugin, and that will remain the case for a few months after `bisect_ppx-ocamlbuild` is released. That will give you a window to switch over to depending on that package. We will announce both the release of `bisect_ppx-ocamlbuild`, and the later point at which it becomes the only way to get the plugin.

I'll revert the docs not to mention `bisect_ppx-ocamlbuild` for now.

aantron:
Ok, instructions should be in sync with reality now. Thanks for asking this.

I'd like to leave this issue open as a reminder to un-revert the instructions during release :)

KasperJanssens:
Alright, thanks for the quick response :-)

PR #153, opened 2017-08-10
Support conditional instrumentation

### LINKED ISSUE #149: Add Jbuilder usage instructions, and enable the instrumentation based on an environment variable
Ok, I was hoping to delay explaining how to use Bisect with Jbuilder until there was a good way to do so (janestreet/jbuilder#57). However, this has been asked for, so we should address it.

We can:

- Enable coverage only if a certain environment variable is present. This is not backwards-compatible, because existing builds expect instrumentation if the PPX runs at all. We could work around this by adding a flag `-conditional` (or something), such that if the flag is present, the PPX is off, unless the environment variable is also set. It's pretty hacky. It should be easy to explain in the docs instructions. It's hopefully temporary, and it's forward-compatible.
- Enable coverage only if an environment variable is absent. This is more backwards-compatible, but seems like a recipe for all sorts of disasters.
- Any other ideas?

cc @kennetpostigo @diml @rgrinberg @lpw25 @andrewray

aantron:
This is a workaround for Jbuilder, which lacks a good way to conditionally disable PPXs. See https://github.com/janestreet/jbuilder/issues/57.

When the option `-conditional` is passed to Bisect_ppx during preprocessing, it checks for the `BISECT_ENABLE` environment variable. If it is present and set to `YES` (case-insensitive), Bisect_ppx instruments code as normal. If it is absent, or not set to `YES`, Bisect_ppx does not modify the AST.

When `-conditional` is not passed, Bisect_ppx instruments as normal. This behavior is backwards-compatible.

A typical `jbuild` file using this would look like this:

```scheme
(jbuild_version 1)

(executable
 ((name foo)
  (preprocess (pps (bisect_ppx -conditional)))))
```

and the corresponding Jbuilder invocation is

```
BISECT_ENABLE=YES jbuilder build foo.exe
```

Usage instructions will be added to the docs when resolving #150.

@let-def, to disable Bisect, I had it return the `default_mapper` to the driver. This implements a "deep identity" mapping. Is a shallow identity mapping readily available?

Resolves #149.

cc @rgrinberg @rleonid @diml @andrewray


rgrinberg:
Looks good, but is the `-conditional` flag really necessary? Why can't the environment variable be just inverted and usable by default? E.g. `BISECT_DISABLE` with the default  being false (unset).

rleonid:
@rgrinberg I think your point is valid, but I do have a slight preference for using the conditional flag.

@aantron Although the cost of the Bisect_ppx runtime is minimal, under the current logic, wouldn't `dump` still get called? Or is there some other logic to disable it in the `jbuilder` ? We still haven't changed the LICENSE, so that would be linking against GPL.

let-def:
@aantron Good idea. I added a shallow_identity (aptly named Ast_mapper.shallow_identity). If this is enough for you I will make a release.

aantron:
@let-def Looks good. If you can make a release within the next couple days, please :) Otherwise, we will pick it up in the next release of Bisect.

@rgrinberg We can invert it, but I think it's better to make instrumentation opt-in than opt-out, although you already have to kind of opt out of it by adding the flag in the first place. I don't want most targets in a `Makefile` to have to be prefixed with `BISECT_DISABLE=yes`. It seems fragile, and not obvious in a project with multiple contributors that this is what needs to be done. What do you think about this?

@rleonid I think it isn't called. When instrumentation is disabled, the runtime module doesn't get referenced, so it doesn't get linked into the binary, even though the package containing it is mentioned as a dependency. The new Jbuilder usage test checks this: https://github.com/aantron/bisect_ppx/pull/153/commits/75057809b26e8b7c514e13f126f1444d7f21af28#diff-a552d9d82f52393c36c01d8ef5d02be8R23

It's a pretty good question, though. Using Bisect_ppx in this way introduces a spurious build-time dependency on it :/ We'll have to document that.

aantron:
@rleonid It does look like we need to protect `dump`. If the binary is linked with `-linkall`, the runtime is linked in, and `dump` is called.

let-def:
@aantron https://github.com/ocaml/opam-repository/pull/10076

aantron:
Thanks!

aantron:
Which brings up the question of *how* to protect `dump`. `BISECT_ENABLE` is meant to be used at build time, not at runtime.

I think the easiest thing to do is just to tell people that this hack is not compatible with `-linkall`.

We could also instrument files when Bisect is off, to set some flag in the runtime not to run `dump`, but I don't want to do that. Or, we could change the behavior of Bisect (when it is on), so that the side effects in the runtime (registering `dump`, reading environment variables) are performed only when triggered by at least one instrumented file. That seems much cleaner.

Any preferences, other ideas?

rleonid:
I understand that you're trying to put lipstick on a pig, so any solution is fine by me.

But honestly, I wouldn't do anything. I don't want to jump through hoops just to accommodate `jbuilder`. `jbuilder` is the _build_ tool, not `bisect_ppx`, it should be able to handle this use case.

aantron:
Yeah, I agree that this should be solved in Jbuilder. Doing this work does not scale to multiple projects that might work like Bisect, among other reasons.

I took (almost) the shortest path possible to making the runtime lazy, and added a new usage test that links a binary with `-linkall`. The test fails if using the non-lazy runtime from `master`, and passes with the one that is now in this PR.

I have a local branch that makes the instrumenter as self-contained and non-side-effecting as possible, so doing something similar to the runtime is not completely out of character, at least :)

PR #155, opened 2017-08-12
Change license to MPL 2.0

### LINKED ISSUE #133: Change the license to something less frightening
We don't believe that the license of Bisect actually affects distribution of any code that uses it, but we would like to avoid hesitation and confusion. My default go-to license is 3-clause BSD, but any other such license (MIT, Apache, etc.) should be good. I believe @xclerc, @rleonid, and myself already agree on doing this; please post which license you prefer or any objections.

aantron:
MPL 2.0 allows Bisect_ppx to be incorporated into proprietary projects, but if it is modified, the modifications must be open-sourced. Contributing them back to Bisect_ppx, or hosting a fork (e.g. on GitHub) is sufficient to meet the requirement.

So, MPL 2.0 is somewhere between BSD/MIT/Apache and LGPL. Besides properietary licenses, MPL 2.0 is compatible with both BSD/MIT/Apache and (L)GPL. See the [MPL FAQ][faq].

See the full discussion in #133. Resolves #133.

I moved the license file to the repo root, so that it can be picked up by GitHub's license detection.

The [suggested file headers][headers] don't include copyright notices. In practice, applying them, as I have done so far in this PR, means that @xclerc's copyright notices are removed. @xclerc, please let me know if you have any objection. My opinion on this is that copyright notices are redundant in a project hosted on GitHub. Some of mine are also removed, but obviously I don't object :)

I left the Ocamlbuild plugin dedicated to the public domain. It seems likely that users would want to customize it, and silly to require those modifications to be open-sourced, or to risk MPL "infecting" `myocamlbuild.ml` that our plugin is inlined into.

cc @rleonid @xclerc @rgrinberg @hcarty @bobzhang

[faq]: https://www.mozilla.org/en-US/MPL/2.0/FAQ/
[headers]: https://www.mozilla.org/en-US/MPL/headers/


aantron:
Rebased to include new files from recent PRs. Merging – I believe we reached agreement in #133.

Issue #114, opened 2016-09-10
documentation for file format


hannesm:
is the file format generated by running the application under test documented somewhere?  if so, I'd appreciate a pointer to that.  thx!

aantron:
It's not documented anywhere, but you can see the reading code for version 1.1.0 [here](https://github.com/aantron/bisect_ppx/blob/1.1.0/src/library/common.ml#L192). `master` [differs](https://github.com/aantron/bisect_ppx/pull/108) somewhat.

Each `.out` file is a [magic identifier](https://github.com/aantron/bisect_ppx/blob/1.1.0/src/library/common.ml#L141) followed by a `Marshal`led `(string * int array) array`, an array of pairs of:

- source filename, and
- array of point visitation counts.

If you want to read this, depending on your environment, the best way might be to link with package `bisect_ppx.runtime` and call `Bisect.Common.read_runtime_data`. Note that I am considering a better interface (#111).

hannesm:
thx! I was hoping to (easily) read it from elisp..  but Marshal is unfortunately very much compiler-version dependent, which makes it a bit hard...

aantron:
We can turn this into an issue for making the file format less dependent on OCaml.

hannesm:
well, if Marshal is convenient for you, I guess I can deal with it (currently low on time in any case, this was just an idea I had days ago -- I dislike having to switch from editor to web browser to view the coverage information)... also, I could try an emacs plugin written in ocaml using janestreet's ecaml

aantron:
It is convenient for Bisect_ppx. It can be annoying to use a format, e.g. JSON, because we are trying to avoid dependencies in Bisect_ppx (since it might complicate the building of those projects, if they ever wanted to use Bisect). But, ultimately it's better to be convenient for Bisect's users, and your idea sounds cool :) So, if you are coming back to this at some point, you are welcome to prod me/others to get it switched. Ideas for a format to avoid dependencies are welcome as well.

If you finish a plugin, let us know and we can link to it.

aantron:
For now, this issue is basically about adding to [`doc/advanced.md`](https://github.com/aantron/bisect_ppx/blob/master/doc/advanced.md) the information at the [top of this issue](https://github.com/aantron/bisect_ppx/issues/114#issuecomment-246787880).

brendanlong:
If you used atdgen to generate JSON, I think it would only be a build dependency.

aantron:
[...]

Issue #111, opened 2016-06-18
New API: summary/meta-issue


aantron:
The Bisect "library" was originally included in the installation only so that instrumented programs could be linked with the runtime. It's an internal interface. However, it has made Bisect somewhat "hackable" ([ocveralls](https://github.com/sagotch/ocveralls), [bisect-summary](https://github.com/lindig/bisect-summary)). That's a good thing :)

To avoid problems like [in #108](https://github.com/aantron/bisect_ppx/pull/108#issuecomment-226267601), and more that are on the way, I think we should try to design a good replacement interface. "Good" means both easy to use, and abstract enough to withstand serious changes in Bisect_ppx.
### Proposal

The interface I can suggest now is something quite similar to @lindig's [proposal](https://github.com/aantron/bisect_ppx/issues/100#issuecomment-222535580). This would replace `Bisect.Common`:

``` ocaml
(* module Bisect.Coverage *)

type location
(* Abstract, see below. *)

val location_to_offset : location -> int
val location_to_line : location -> int

type point =
  {location : location;
   hits     : int}

type file =
  {points      : point array;
   line_count  : int;
   source_code : string option}

type t = (string, file) Hashtbl.t

val read : ?source_code_directory:string -> ~out_files:string list -> coverage
```

At a file system level, there would be only `.out` files. They would be self-contained. They would include everything now in `.out` files, `.cmp` files (#102), and the source code (#103). So, a tool would only need to find a collection of them, and pass their names to `read`. This should make both programmatic use and use of `bisect-ppx-report` easier – especially in contexts where there isn't access to the build directories.
### Details
- Lines are included in locations, so that it's not necessary to walk over the source to convert offsets to lines. Also, the source may be missing.
- Type `location` is abstract, because we might need to switch to location ranges to address a [potential future problem](https://github.com/aantron/bisect_ppx/issues/82#issuecomment-226895380):
  
  > There may be some challenges, when an expression has the same location as some relevant subexpression.
  
  I'd also like the freedom to decide between using `Location.t`, and thus depending on `compiler-libs`, or not.
- A digest can be easily computed from `source_code`. We can include the digest, so that ocveralls can work even if the source is missing, but I'm wondering if it's not better for ocveralls to simply fail in that case. @lindig, what was your reason for including the digest?
- `line_count` is there for the same reason as lines in locations, so that you can know how many lines to report as unvisited without scanning the source, especially if it is missing.
[...]

PR #165, opened 2017-12-23
CSV report with per-line coverage info

brendanlong:
A lot of tools don't support coverage by file offset, so this is a new report that reports coverage by line.

Due to the fact that a line can have both covered and uncovered parts, this shows both pieces of info and lets downstream tools figure out what to do with that.

As an example of where this would be useful, [Phabricator's coverage reports](https://secure.phabricator.com/book/phabricator/article/arcanist_coverage/) only understands per-line "covered", "uncovered", "not executable" and "unreachable. With this new report, we can show any line with an unvisited point as "uncovered", any line with a visited point that doesn't also have a visited point as "covered", and anything else as "not executable".

### DIFF
--- a/src/report/report.ml
+++ b/src/report/report.ml
@@ -7,6 +7,7 @@
 type output_kind =
   | Html_output of string
   | Csv_output of string
+  | Csv_line_output of string
   | Text_output of string
   | Dump_output of string
 
@@ -74,6 +75,10 @@ let options = Arg.align [
   ("-csv",
    Arg.String (fun s -> add_output (Csv_output s)),
    "<file>  Output CSV report to <file>");
+
+  ("-csv-line",
+   Arg.String (fun s -> add_output (Csv_line_output s)),
+   "<file>  Output line coverage CSV report to <file>");
 
   ("-separator",
    Arg.Set_string csv_separator,
@@ -173,6 +178,9 @@ let main () =
           search_in_path data points
     | Csv_output file ->
         generic_output file (Report_csv.make !csv_separator)
+    | Csv_line_output file ->
+        Report_csv_line.output verbose file !csv_separator search_in_path data
+          points
     | Text_output file ->
         generic_output file (Report_text.make !summary_only)
     | Dump_output file ->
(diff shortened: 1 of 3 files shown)


brendanlong:
@bradlangel @kevinqiu I based this pull request on your code in https://github.com/arenadotio/bisect_ppx/commits/line_coverage_latest

I tried to simplify it a little bit by not making decisions about what visited + unvisited should mean, and printing one CSV line per line instead of trying to combine it so each line represents an entire file.

brendanlong:
Here's an example of the output (from [pgx](https://github.com/arenadotio/pgx)):

brendanlong:
And I still have mixed feelings about whether this should be in bisect_ppx or not -- I'm looking at this the other way around and also writing a Phabricator extension that reads the normal CSV and the files and figures this info out externally. I figured other people might want something similar so it was worth a try though.

brendanlong:
An alternative would be having some sort of output we can easily access from other languages.

aantron:
References to other issues first:

1. About documenting the current file format, or providing JSON output that can easily be used by other languages: #114.
2. About including line numbers and not just offsets in location information, so that external tools don't have to compute them: #111.

We can merge this PR, but could you please glance and 1, 2? If it's not an emergency to merge, and those two issues would solve the problem more generally and conclusively, perhaps I should just commit to doing a couple 2.0.0 issues per week until that release is ready :)

brendanlong:
I can look into #114 and #111 isntead. I just made our internal thing parse the `-dump` output, which is.. not ideal, but it works for now.

aantron:
Yeah, all of that should be made much easier, and external tools shouldn't have to re-parse source files.

If the stuff described in #111 especially seems good, let me know, and I'll work on it. You're welcome to do it as well, of course :)

brendanlong:
I've been thinking about this more, and I think what I should really do is make a separate project like https://github.com/sagotch/ocveralls to generate a Phabricator report. Or just use it's output JSON since it's close to what we want anyway.

Having a JSON version of the `dump` output would still useful but I'm not sure it's worth it to have this alternate CSV output.

aantron:
That seems reasonable.

I think in the long term, we want to move some of what ocveralls does into Bisect_ppx, though (like the line number computation).

brendanlong:
Feel free to take anything from this pull request that's useful, but I'm guessing the CSV isn't that helpful in general so I'll close this.

aantron:
Ok, thank you :)

Issue #82, opened 2016-02-10
Instrument correctly


aantron:
Reported by David Mentre on the mailing list: https://sympa.inria.fr/sympa/arc/caml-list/2016-02/msg00038.html

aantron:
Bisect_ppx isn't disciplined about how it places points. Consider:

![screen shot 2016-06-17 at 17 11 22](https://cloud.githubusercontent.com/assets/12073668/16166189/a91dcb66-34ae-11e6-84c4-625f0017cda1.png)

The point apparently on `later` in `everything` is actually on `sooner`, since that has to be evaluated first. There is no point that would indicate that `sooner` completed and `later` was reached. In `everything_sequence`, both places have a point.

I think the rationale for putting points on sequence expressions was that if an exception occurs in the first, the second is not reached. However, similar statements are true of most constructs in the language that have subexpressions.

Now that we don't have point kinds, we should consider putting points on many more expressions, and doing it correctly.

There may be some challenges, when an expression has the same location as some relevant subexpression. I can't immediately think of an example, however.

I think the instrumenter's mapper will probably need a rewrite to address all expression kinds correctly.

PR #166, opened 2017-12-23
Add a JSON report

brendanlong:
This outputs something like:

```
{
  "file.ml": {
    "points": [{
      "offset": 0,
      "identifier: 5
    }, {
      "offset": 10,
      "identifier": 0
    }]
  },
  "file2.ml": {
    ..
  },
  ...
}
```

This is basically the same as the `dump` format, but
much easier to parse.

The per-file report is an object containing the `"points"`
key rather than just the points themselves, since this
will allow us to add other keys in the future without
breaking backwards compatibility (for example, to add per-line
output).

Downsides:

 - This brings in a dependency on atdgen and yojson
 - Atdgen's generated .ml files have compile warnings,
   so we can't build with `--dev`. I'm not sure how to
   work around this yet.


brendanlong:
I'm not sure if it's worth it to pull in atdgen + yojson, plus having the compiler warnings, but here's a start on this.

It's possible that we should just use yojson without atdgen to minimize the dependencies and not have to deal with compiler warning problems.

brendanlong:
Here's an example report (from pgx again): https://pastebin.com/Yz5FDzau

aantron:
We could either disable the warning that is being emitted, or generate JSON manually. I think even if we merge this with an `atdgen` or `yojson` dependency, we will remove it before release.

brendanlong:
Ok, this removes the atdgen dependency but leaves yojson. I could probably make it write JSON manually but I don't know if we want to go down that path?

brendanlong:
I'm starting to wonder if this should just be a separate binary that's not part of bisect_ppx... but I don't really want to maintain that :\

aantron:
It should probably be part of Bisect_ppx, though we're happy to make Bisect "hackable" by making it easy to write extra binaries when needed, e.g. [ocveralls](https://github.com/sagotch/ocveralls), [bisect-summary](https://github.com/lindig/bisect-summary).

I would prefer manual JSON generation, but it's not a big deal to depend on Yojson. We can always make the decision about the dependency later, before the next release, especially if we add tests for the report output (e.g. https://github.com/aantron/bisect_ppx/tree/master/test/unit/fixtures/report).

As mentioned elsewhere, I started doing "serious" work on Bisect_ppx again, slowly, starting with fixing #82. I think the points format is going to change as a result, but, again, we can stabilize that before next release.

And if you think a separate binary is the best for your needs now, or forever, we're happy to support you on the Bisect side :)

Issue #169, opened 2018-04-19
presence of .exclude file should not matter when bisect is disabled


gtrak:
This is a multi-project interaction, but I think bisect_ppx's responsibility is to simply not break when it's not actually doing any work.

What's expected to happen:
Bisect and jbuilder play nicely, and paths are properly sorted out so that merlin can work.

What happens:
merlin is run without any knowledge of your project root, so absolute paths are required in .merlin.  There is proof of the issue with ppx drivers themselves and a hard-coded solution for that case: https://github.com/ocaml/merlin/issues/699

But this is not enough when the ppx requires external configuration files. 

The external effects with merlin is to show errors like 'extension 'map' not supported'.

I have filed a corresponding jbuilder issue to be able to absolutize the location of the .exclude file: ocaml/dune#712

It seems to me that bisect is not needed during development time when it's being used conditionally, therefore it shouldn't break if configured improperly at that time.  Can we add a check or early-exit or something that avoids this problem?

Here's how it's being invoked in our jbuild:
```
  (preprocess (pps (ppx_jane bisect_ppx -conditional
                    (-exclude-file email_report/src/.exclude))))
```

And I would like the behavior when the bisect environment variable is unset to just pass things through regardless of configuration.

gtrak:
Thoughts? https://github.com/arenadotio/bisect_ppx/pull/1/files

gtrak:
I verified that the fix works and merlin is fine now.

aantron:
Thanks for reporting :) The fix looks good, would you mind sending it as a PR against this repo as well?

gtrak:
sure!

PR #170, opened 2018-04-19
Make adding exclude files conditional

gtrak:
Fixes aantron/bisect_ppx#169

### DIFF
--- a/src/ppx/register.ml
+++ b/src/ppx/register.ml
@@ -6,13 +6,31 @@
 
 let conditional = ref false
 
+let enabled () =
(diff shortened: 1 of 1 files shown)


aantron:
Thank you, will merge after CI :)

aantron:
Thanks again :) Would you like a speedy release to opam?

gtrak:
We've got a fork opam pin working already, and I haven't tested the other changes since 1.32 so I'm not in a rush :-).  Just needed to unbreak my emacs this morning.

PR #171, opened 2018-04-28
Refactor file registration

hhugo:
The main side effect is that one can now use bisect_ppx in context where `Array` & `Pervasives` are not the one from the compiler stdlib.


aantron:
Thank you.

Looking at the linked PR, I guess a speedy release of this to opam would be helpful. Please let me know if that's the case.

Issue #193, opened 2019-05-22
change output file creation?


yomimono:
I was experimenting with combining bisect_ppx with the AFL instrumentation in the compiler.  It's surprisingly non-horrible, except for one thing: the program can easily be run tens of thousands of times during one test run, and so the logic by which Bisect [figures out where to write its output](https://github.com/aantron/bisect_ppx/blob/master/src/runtime/runtime.ml#L53) becomes increasingly slow.

I have a patch which just randomizes the number suffix, which I'm happy to submit if you find that to be an acceptable change.  Otherwise I'm happy to submit a patch with different behavior if you have some intuition for what better behavior might be.

aantron:
Randomization sounds like a good idea, the patch would be welcome :)

rleonid:
@aantron Is the boss here (I'm just a lurker still getting notifications) and I like randomization as much as the next dev... but maybe a more robust algorithm would be to switch to a directory scanning after 3-5 tries, or add an environment variable to guide behavior? `Unix.opendir`, `Unix.readdir` and grab last bisect file and +1 ?

aantron:
I'd say randomization is the easiest thing to do and understand. If @yomimono has already observed the patch working and solving the problem, I would prefer to merge that.

I hope nobody is relying on `bisect0001.out` being created with exactly that number :) But if so, Bisect is not production anyway, so this is just a small annoyance for a few developers to replace with `bisect*.out`.

yomimono:
PR'd over at #194. It's a super small change but I chose to dig it out of `git reflog`, then had to learn how to do that, hence the delay :sweat_smile:

PR #194, opened 2019-05-28
randomize output file name

yomimono:
Instead of incrementally trying file names from `bisect_0001.out` and up, randomize the number suffix.  This is intended to keep the 10,000th consecutive run of a program from looking for 9,999 files before writing, with motivation further described in #193 .


aantron:
Thanks, I opened #195 about the error in CI, and I'll fix that. I'll rebase and merge this afterwards. This could take a day or more, but you shouldn't have to do anything further, so consider it notionally merged :)

aantron:
Thanks again!

aantron:
I think we also need to call `Random.self_init ()` to make this work. Did you have this in your full change? I will add a patch.

PR #198, opened 2019-07-14
Control instrumentation using AST attributes

### LINKED ISSUE #83: Look into using attributes to control points
Suggested by Alain Frisch: https://sympa.inria.fr/sympa/arc/caml-list/2016-02/msg00044.html

This would be in addition to comments, not instead – there doesn't seem to be a reason to break backwards compatibility.

### LINKED ISSUE #196: Type declarations marked as unvisited
I've just integrated this awesome ppx into my Esy/Reason/Dune project, and looking at the coverage report I noticed that bodies of type definitions are not marked as visited, decreasing file coverage. Is that expected?

image

aantron:
This PR adds three kinds of `[@coverage ...]` attributes:

1. `[@@@coverage off]` and `[@@@coverage on]` for disabling coverage instrumentation inside all expressions in a block of structure items:

    ```ocaml
    [@@@coverage off]

    let foo = ...

    let bar = ...

    [@@@coverage on]
    ```

    @Lupus, in particular, if you surround a type definition that has `[@@deriving sexp]` with these, coverage is suppressed for the generated helpers. At least it was in my testing :)

2. `[@@coverage off]`, which applies to a single binding in a `let`...`and` structure item, and all of its subexpressions:

    ```ocaml
    let foo = ()
      [@@coverage off]
    ```

3. `[@coverage off]`, which applies to an expression and all its subexpressions:

    ```ocaml
    if a then
      f x [@coverage off]
    else
      g y
    ```

Of course, in Reason, they all have the syntax `[@coverage off]`/`[@coverage on]`, with only one `@`, and have to be positioned differently and correctly.

I'll add deprecation warnings to the comments in separate work, and remove them in some future release.

Resolves #196.
Resolves #83.

### DIFF
--- a/doc/advanced.md
+++ b/doc/advanced.md
@@ -13,6 +13,7 @@ them after `bisect_ppx` in the `pps` list.
   - [With Ocamlbuild](#Ocamlbuild)
   - [With OASIS](#OASIS)
 - [Excluding code from coverage](#Excluding)
+  - [Expressions and structure items](#ExcludingExpressions)
   - [Individual lines and line ranges](#ExcludingLines)
   - [Files and top-level values](#ExcludingValues)
 - [Environment variables](#EnvironmentVariables)
@@ -209,8 +210,23 @@ The easiest way to exclude a file from coverage is simply not to build it with
 need finer control. There are several ways to disable coverage analysis for
 portions of code.
 
+<a id="ExcludingExpressions"></a>
+#### Expressions and structure items
+
+You can tag expressions with `[@coverage off]`, and neither they, nor their
+subexpressions, will be instrumented by Bisect_ppx.
+
+Likewise, you can tag module-level `let`-declarations with `[@@coverage off]`,
+and they won't be instrumented.
+
+Finally, you can turn off instrumentation for blocks of declarations inside a
+module with `[@@@coverage off]` and `[@@@coverage on]`.
+
 <a id="ExcludingLines"></a>
 #### Individual lines and line ranges
+
+*Note: this method is deprecated and will be removed; use the approach
+described above in "Expressions and structure items."*
 
 If a comment `(*BISECT-IGNORE*)` is found on a line, that line is excluded from
 coverage analysis. If `(*BISECT-VISIT*)` is found, all points on that line are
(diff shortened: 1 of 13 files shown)


Lupus:
Sounds perfect! I'll give it a try once I get back to the project which I ran into this :) :+1:

aantron:
Great :) I will merge this to avoid it going stale, and we can make further changes later, if they are needed.

Lupus:
I've got back to this! Works good, but it seems that it's impossible to exclude a branch of match from coverage :( I have a bunch of obvious branches with failwith asserting for nearly imaginary error conditions so that it is unfeasible to cover those with tests. Am I missing something?

Lupus:
image

Also I'm a bit lost as of why this does not work...

aantron:
Can you confirm the version/commit you are using, and show the syntax you are trying for excluding a match case? Also, can you show what worked, so I can compare it to the screenshot that shows the attributes not working?

aantron:
There is also #246 for being able to instrument match cases themselves, as opposed to their expressions.

Lupus:
That's what I have in my switch:
```
$ opam list | grep bisect_ppx
bisect_ppx              1.4.1       Code coverage for OCaml
```

This does not work either:
image

Am on on a wrong version? :)

This was my naive attempt at excluding whole case from coverage:

image

aantron:
This PR has been merged into `master`, but not released, and in any case it was done after 1.4.1. So if you really have 1.4.1, and not a pin, then none of these attributes will do anything. What do you have when you do `opam pin list`?

`[@coverage off]` in patterns is not supported yet even in `master`, #246 is still open :)

Lupus:
Got it, for some reason I got the impression that I've got preempted to another project for sufficient amount of time for a new release to magically appear :joy: So I don't have any pinnings and the fact that nothing happens is very well expected.

I'll recheck the condition next time I got awaken by coverage conditional variable so as not to clutter your inbox. Sorry for the noise.

P.S.: coverage in patterns sounds like a nice idea actually!

aantron:
No problem :)

Bisect_ppx is actually almost ready for release from a technical point of view. I am the limiting factor at the moment :/

Lupus:
That's fine, we're all busy with lots of stuff on our plates :)

Issue #267, opened 2020-01-19
Environment variable to toggle the PPX


vphantom:
Related to #203, maybe this project could do like [Landmarks](https://github.com/LexiFi/landmarks) with its `$OCAML_LANDMARKS` environment variable.   Users could leave `bisect_ppx` in `dune` files at all times and, to get a coverage-enabled build:

```sh
dune clean ; OCAML_BISECT=auto dune build @all
```

Then for a production release, without the environment variable the PPX would stand down and not inject anything at all:

```sh
dune clean ; dune build @all
```

It's fairly clean and doesn't require any change in Dune. The only drawback is that projects structured this way would depend on `bisect_ppx`, which for my use is an acceptable compromise since it doesn't affect the compiled output in release mode.

aantron:
@vphantom If I understood correctly, this is already what [`BISECT_ENABLE`](https://github.com/aantron/bisect_ppx#dune) does.

aantron:
Perhaps the README should clarify the variable's exact effect. IIRC we had that in an earlier version of the docs, but it got lost somewhere in an effort to simplify them.

vphantom:
Thank you for clarifying.  What confused me was that step 4 of the Dune section of the README mentions removing the `(preprocess ...)` rule from the `dune` file for a production release (awaiting its issue 57).

Going down in the Real-world example section, I now see that indeed, `--conditional` tells the PPX to pass through untouched when the `BISECT_ENABLE` environment variable isn't set.

I'll attempt a quick PR to reword this.

PR #268, opened 2020-01-20
Update README's Dune section re: $BISECT_ENABLE

vphantom:
Per #267, my understanding is that the mention of a limitation in Dune has been worked around with `--conditional` and setting `BISECT_ENABLE=yes` at build time (in addition to run time for generating the report).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,9 @@ Refer to [**aantron/bisect-starter-dune**][dune-repo], which produces
      (public_name my_lib)
      (preprocess (pps bisect_ppx --conditional)))
     ```
+    ```sh
+    dune clean ; BISECT_ENABLE=yes dune build @all
+    ```
 
 3. Run your test binary. In addition to testing your code, when exiting, it will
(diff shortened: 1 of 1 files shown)


aantron:
I committed an alternative clarification in https://github.com/aantron/bisect_ppx/commit/a2a91a3e76d44b4a4dd24d529482ba9e2ef9c37c. The main reason is that many projects don't want a dependency on Bisect when releasing. In particular, this is common for libraries, since they don't want to introduce an unnecessary transitive dependency to *their* users.

The clarification does offer instructions in case a project is ok with leaving Bisect in.

I also more explicitly clarified the purpose of `BISECT_ENABLE=yes` and `dune clean`.

vphantom:
Excellent!

I never used GitHub's PR process this extensively; I just now clicked "Merge" and favored your version, which seems to have satisfied it. I hope this was appropriate.

aantron:
I think that updated this PR, so no problem :) I think the PR can be closed.

PR #288, opened 2020-03-20
fix generated html

zapashcanon:
Hi,

I just fixed the generated HTML. I also added `lang="en"` as the title is in English and the code is likely to be.

Btw, when I tried to do `dune runtest` something is asking me to unlock my PGP key. I guess it has something to do with the CI, but, isn't there a way to run them locally ?

Cheers.

### DIFF
--- a/src/report/report_html.ml
+++ b/src/report/report_html.ml
@@ -54,7 +54,8 @@ let output_html_index verbose title filename l =
     filename
     (fun channel ->
       Report_utils.output_strings
-        [  "<html>" ;
+        [  "<!DOCTYPE html>" ;
+           "<html lang=\"en\">" ;
            "  <head>" ;
            "    <title>$(title)</title>" ;
            "    <link rel=\"stylesheet\" type=\"text/css\" href=\"coverage.css\" />" ;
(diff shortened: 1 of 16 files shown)


aantron:
I run the tests locally without this prompt. I suggest to do a binary search for which test is causing this, by commenting out test suites, and then test cases within a test suite. The "root" is here:

https://github.com/aantron/bisect_ppx/blob/c96bef058359f2dc3a5c6eaaf0515319774da46e/test/unit/test_main.ml#L9-L25

aantron:
Would you be willing to also fix the HTML for per-file pages?

https://github.com/aantron/bisect_ppx/blob/c96bef058359f2dc3a5c6eaaf0515319774da46e/src/report/report_html.ml#L217

aantron:
If you do, it should cause a diff in one of the reporter tests. You can promote the output after reviewing it with `make save-test-output`, then run the tests again.

zapashcanon:
> Would you be willing to also fix the HTML for per-file pages?

Done in the first commit.

> If you do, it should cause a diff in one of the reporter tests. You can promote the output after reviewing it with make save-test-output, then run the tests again.

I did something in the second commit. Not sure if it's correct...

aantron:
Thanks!

When I ran the tests on my machine, I had to revert the diff in the `send-to` test output. That's also, I think, what would have happened in Travis - not sure why it didn't run. I suppose this is related to the issue you had in the first comment, but I don't know what the problem is/was.

PR #289, opened 2020-03-21
add link to routes codecov

anuragsoni:
Just realized that bisect supports github actions :smile:


aantron:
Thanks! Glad to see this working "in the wild" :)

PR #296, opened 2020-03-25
Fix dune guidline for OCaml < 4.07

kit-ty-kate:
Issue encountered in https://github.com/ocaml/opam-repository/pull/16066

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,11 +67,10 @@ Refer to [**aantron/bisect-starter-dune**][dune-repo], which produces
     (* -*- tuareg -*- *)
 
     let preprocess =
-      try
-        match Sys.getenv "BISECT_ENABLE" with
-        | "yes" -> "(preprocess (pps bisect_ppx))"
-        | _ -> ""
-      with _ -> ""
+      match Sys.getenv "BISECT_ENABLE" with
+      | "yes" -> "(preprocess (pps bisect_ppx))"
+      | _ -> ""
+      | exception Not_found -> ""
 
     let () = Jbuild_plugin.V1.send @@ {|
 
--- a/README.md
+++ b/README.md
@@ -67,9 +67,11 @@ Refer to [**aantron/bisect-starter-dune**][dune-repo], which produces
     (* -*- tuareg -*- *)
 
     let preprocess =
-      match Sys.getenv "BISECT_ENABLE" with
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,9 +67,11 @@ Refer to [**aantron/bisect-starter-dune**][dune-repo], which produces
...
 
     let preprocess =
-      match Sys.getenv "BISECT_ENABLE" with
-      | "yes" -> "(preprocess (pps bisect_ppx))"
-      | _ | exception _ -> ""
+      try
+        match Sys.getenv "BISECT_ENABLE" with
+        | "yes" -> "(preprocess (pps bisect_ppx))"
+        | _ -> ""
+      with _ -> ""
```
aantron: Would a separate `| exception Invalid_argument _ ->` case (not an or-pattern) still work, and be more tidy?

kit-ty-kate: I'm not sure what you're suggesting. The issue that this PR is fixing is that `| exception e ->` is only valid since OCaml 4.07. This pattern did not exist before.

Otherwise feel free to change my code if there is a coding style that you prefer.

aantron: `| exception e -> ...` was added in [4.02](https://caml.inria.fr/pub/docs/manual-ocaml/patterns.html#sss:exception-match). What was not allowed (it seems) is using it inside an or-pattern, as opposed to at the top level of a case. So `match foo with | exception e -> ...` was ok, but `match foo with | p | exception e -> ...` was not.

aantron:
Thanks!

aantron:
(Tested the modified code on 4.02 in the starter repo).

This PR actually raises the natural question of how Bisect's instrumenter handles `exception` that is not at the top level, so I opened #297.

aantron:
Amazingly, both AppVeyor and Travis seem to be having problems right now (not due to this repo). Sorry about the CI mess.

README.md
kit-ty-kate: Ohhh I thought it was added much later. Sorry ^^"

aantron: Meanwhile, I had always assumed it was only allowed at the top level since 4.02, tried it in an or-pattern on a whim for the README, and failed to realize that Bisect_ppx itself doesnt handle this extension :p

Issue #164, opened 2017-12-19
Interaction with other ppx rewriters in 1.3.1


emillon:
Hello,

I hit a regression with `bisect_ppx`. 1.2.0 works, but when using 1.3.1 we now have a build error.

I reduced the following test case:

```
# _tags
true: package(ppx_deriving.std)
true: package(ppx_bin_prot)
true: package(bisect_ppx)
```

```
(* test.ml *)
module type EQ = sig
  type t [@@deriving eq]
end
```

```
(* test.mli *)
module type EQ = sig
  type t [@@deriving eq]
end
```

(build with `ocamlbuild -clean && ocamlbuild -use-ocamlfind test.byte`)

Using `bisect_ppx.1.3.0`, `ppx_deriving.4.2.1` and `ppx_bin_prot.v0.9.0`, this fails with the following error:

```
File "_none_", line 1:
Error: eval expected
Command exited with code 2.
```

Some remarks:
- removing the `package(bisect_ppx)` line fixes it
- removing the `package(ppx_bin_prot)` changes the error to `Unrecognized [@@deriving] annotation syntax` (note that this module does not use it it)
- removing the interface file fixes it
- this works with 1.2.0 and not with 1.3.1. I bisected it down to 8fe184ccf38e8c7c2da270ad6774d0f5aee338f6.

It seems that it's a new bug in the exclusion mechanism, maybe triggered by strange locations emitted by these ppx rewriters.

Let me know if I can be of further help.

Thanks!

aantron:
Thanks for the report. I saw what is probably another symptom of this problem. It looked like in some cases, Bisect_ppx would write its per-module instrumentation initialization code into an attribute of the module, rather than the module itself. This is probably what is causing the `eval expected` message above, as when Bisect does that, it prepends other structure items before the expression-evaluation structure item that was originally in the attribute.

Other symptoms were losing documentation, when the attribute was an ocamldoc string, and not generating any instrumentation output, as the code written into the attribute doesn't ever run.

I couldn't reproduce it after twiddling my opam switch, though, so I didn't fix it. Hopefully, I'll be able to track it down firmly now. I'll take a look in some hours or maybe days.

aantron:
I just realized what the bug almost certainly is. Bisect_ppx chooses which structure to insert its initialization code into by inserting it into the first structure that it encounters while traversing the AST.

In an `.ml` file, that first structure is always the one contained in the module representing the whole file.

[...]

PR #327, opened 2020-07-22
Port the ppx from omp to ppxlib

pitag-ha:
Before, the ppx was directly based on omp: on the omp driver and on omp's Ast_411. Both of them will get dropped in omp 2.0.0. For more information on that omp change, see https://discuss.ocaml.org/t/ocaml-migrate-parsetree-2-0-0/5991 by @jeremiedimino. 

That's why this commit ports the ppx from omp and ppx_tools_versioned to ppxlib.
Closes https://github.com/ocaml-ppx/ppxlib/issues/131


pitag-ha:
I have a question about the Bucklescript support: I'm not familiar with Bucklescript and it's divergence from OCaml. Does the support here still work? Is it worth making ppxlib be able to keep it?

aantron:
BuckleScript support should work with this approach, as for BuckleScript support, we build the Bisect_ppx binaries using the opam ecoystem (though using esy as the package manager). However, it is broken in this PR at the moment: https://travis-ci.org/github/aantron/bisect_ppx/jobs/710719501.

Travis seems to not be reporting status to GitHub (a problem they have been having on and off for months now). The builds are running, however. They are at https://travis-ci.org/github/aantron/bisect_ppx.

aantron:
I see that this makes Bisect_ppx transitively depend on Base. From discussions with @jeremiedimino and @rgrinberg, I got the impression that Base would eventually not be a dependency of ppxlib or other future PPX helper library. Is that still the case? I strongly prefer not to introduce Base as a dependency of Bisect_ppx.

jeremiedimino:
@aantron the depencency on Base has been dropped in the master of ppxlib. The next release will not depend on Base.

pitag-ha:
Thanks for the answer @aantron and for pointing me to the failing tests. We will integrate the BuckleScript support into ppxlib and once that's done apply it here.

I've been trying to figure out why the BuckleScript tests fail, but haven't found the connection to what I've ported, yet. I haven't touched that part of the code. For example it says that the `-conditional` option is not found. That part is still using the omp driver. And that ""make -C test/bucklescript install" failed. I'm not sure how I might have influenced that. Do you happen to have an idea how I can have triggered those errors by only porting the OCaml ppx?

I will leave the PR marked as WIP until the next ppxlib release to avoid the Base dependency.

aantron:
I'm not familiar enough with ppxlib to immediately say what caused that error. I did [rerun](https://travis-ci.org/github/aantron/bisect_ppx/builds/706868946) the CI of `master` to (apparently) rule out changes upstream as the cause.

pitag-ha:
I hope the failure of the bucklescript tests will get solved once ppxlib has an integrated bucklescript support. I have a couple of bucklescript unrelated questions in the meanwhile if you don't mind: I don't understand the [signature method](https://github.com/aantron/bisect_ppx/blob/master/src/ppx/instrument.ml#L1500-L1503) of the instrumentor class. I have two questions about that:
1. In the [comments it says](https://github.com/aantron/bisect_ppx/blob/master/src/ppx/instrument.ml#L1490-L1491) that the signature method will only set a variable to true and do nothing else. But after setting that variable to true, it [recurses through the AST](https://github.com/aantron/bisect_ppx/blob/master/src/ppx/instrument.ml#L1503). Is that intentional?
2. Why is it necessary to have that signature method? Is it a problem to insert the Bisect_ppx initialization code into a struct of an interface file? As far as I've been told, it would just get ignored by the compiler. Are there cases where that's not true?

I'm just asking because [I've implemented the possibility to register a ppx as an instrumentor](https://github.com/ocaml-ppx/ppxlib/pull/161) in ppxlib (which allows one to specify if one wants the instrumentor transformation to be applied before or after the other transformations). But that implementation assumes that an instrumentor only passes through implementation files.

aantron:
Blame shows the code was added in #164, so:

> Is it a problem to insert the Bisect_ppx initialization code into a struct of an interface file?

It was a problem in the past, in #164. Bisect_ppx would insert initialization code into structures that appeared inside attributes in interface files.

> As far as I've been told, it would just get ignored by the compiler. Are there cases where that's not true?

Indeed, it should get ignored by the compiler, but attribute payloads are interpreted by other PPXs, and inserting Bisect_ppx initialization code into them would interfere with those PPXs, as happened in #164. 

This code might be obsolete, because we now have an additional way of not instrumenting attributes:

https://github.com/aantron/bisect_ppx/blob/a4b8cb617dd9b8f33be8da7b8e63390e81b85fb6/src/ppx/instrument.ml#L1478-L1483

It probably got left in as a form of extra defensive programming. During and after #164, Bisect_ppx still had some convoluted way of avoiding attributes that evidently wasn't working.

If you intend to avoid applying Bisect to interface files by other means, this method would become even more redundant.

The recursive call in the method probably dates from the time when Bisect_ppx was still trying to do something fancy with attributes, since, to my knowledge, attributes are still the only way "instrumentable" code can appear inside a signature (and we don't instrument it now anyway).

pitag-ha:
Thanks a lot for that detailed answer! I've avoided applying Bisect to interface files via the ppxlib API now. I've also used that API to do the work that was done by the `saw_top_level_structure_or_signature` variable before . Please, let me know if you don't like that change. I don't have any problem in reverting it.

src/ppx/instrument.ml
```diff
@@ -1482,75 +1470,58 @@ class instrumenter =
...
+          Coverage_attributes.has_exclude_file_attribute ast
+        in
+
+        let get_path = function
+          | [] -> None
+          | hd :: _ -> Some hd.pstr_loc.loc_start.pos_fname
+        in
+
+        match get_path ast with
+          | None -> ast
```
aantron, line 1511: Under what condition(s) will this be `None`? I suspect that we want to instrument in those cases, if they are possible.

binaries.esy.json
```diff
@@ -1,10 +1,9 @@
 {
   "dependencies": {
-    "ocaml": ">= 4.2.0",
+    "ocaml": ">= 4.4.0",
```
aantron, line 3: Is this properly a constraint of Bisect? Is there anything added by this PR to Bisect itself, that requires OCaml 4.4.0? The only candidate I see are the changes to the labeled argument representation.

If not, I think we should rely on ppxlib's constraints.

On the subject of those, I would weakly prefer if ppxlib supported 4.02. It's probably not so critical for Bisect, as a typical project can just run coverage on a newer version of the compiler.

aantron (commented):
The code looks good, with a couple comments.

Please don't force-push into this PR until the end, it complicates incremental review :) We can rewrite the history as necessary at the very end, probably with just a squash-merge.

src/ppx/instrument.ml
pitag-ha: This will be `None` whenever the structure given to the rewriter is empty. Two cases I know of in which the structure is empty:
- When having an empty .ml file, the structure corresponding to that file is empty.
- When loading a toplevel loading a ppx deriver (as done for example in `test_top_level` test), the deriver will get loaded and applied to an empty structure before getting loaded and getting applied to the input. To be honest, I don't know why it's like that.

binaries.esy.json
pitag-ha: Yes, you're right, this is not needed here. I've gone back to the 4.2 bound.

I will mention that you'd prefer a 4.2 support for ppxlib in the next ppxlib meeting. But we've already talked about supporting 4.2 once and, I think, unless it's necessary, it's likely that we'll have to focus on other things.

pitag-ha:
I had to exclude the `test/usage/reason` directory from the dune build, since the ppxlib version I need has a conflict with `reason`. Is that ok? That problem has always been there, but I've noticed it now after merging your dune2 integration.

I've also commented out the bucklescript integration, since it's not working the way it is now. I think that's why the CI is failing now. I'll fix it as soon as ppxlib supports bucklescript.

src/ppx/instrument.ml
pitag-ha: Is it ok to leave it the way it is? Or would you like me to find out all possible cases in which the structure is empty?

pitag-ha: I've recently understood that the way I was trying to get the source filename might be problematic, since `hd.pstr_loc` might be `Location.none`, for example if a ppx runs before `bisect_ppx` using `Location.none`.  I've also just seen that the source filename is not only used to decide if to traverse the AST or not, but also inside the runtime initialization. So it doesn't seem like a good idea to try to get around using `Location.input_name` here. I've gone back to using it, even though that means relying on compiler-libs here. If ppxlib introduces a similar concept in the near future or I find another way around it, I'll adapt the PR.

aantron:
Thanks, that all sounds reasonable. It's ok to disable the Reason test. Hopefully, we can enable it again after some time and a few releases of the upstream projects.

aantron:
I think Bisect needs this PR for proper, maintainable 4.12 support. What's the status of the PR and ppxlib now?

pitag-ha:
I'm integrating the possibility into ppxlib to access the `input_name`, so that we don't need to use compiler-libs for that. That's almost done. Afterwards I need to implement bucklescript support for ppxlib. I don't know yet how much work that will be. But knowing that this is needed for 4.12 support I'll prioritize ppxlib and this PR and try to go faster.

pitag-ha:
@aantron , can I rebase over master and force-push now to see if the CI passes? I've fixed the bucklescript support in Ppxlib and made the input name accessible via Ppxlib in whole file transformations. And I've adapted the PR to those two changes.

aantron:
@pitag-ha Thanks. I suggest, instead of rebasing, to merge `master` into this branch. If you have a strong reason specifically to rebase, however, then go ahead.

pitag-ha:
Ok, I've merged master into this branch. I'm having two problems now:

- In my last commit, I changed the version of ocamlformat from 0.15.0 to 0.15.1 to have it compatible with omp2. But I've just noticed that you're using a binary from your fork. Would it be possible to add ocamlformat 0.15.1 or 0.16.0 to your fork to have omp2 compatibility?

- There's one test that's failing locally at the moment (`test/instrument/class/`). It's also failing in master.

pitag-ha:
And a third problem: the bucklescript test installation fails on the CI, so I can't see if the bucklescript test would pass or not (locally it did). Do you happen to know what fails in the installation?

src/ppx/instrument.ml
```diff
@@ -32,26 +32,14 @@
    point is in [bisect_ppx.ml]. It's basically a PPX driver that registers only
    this instrumenter with itself, using [register.ml], and then runs it. *)
 
+open Ppxlib
```
aantron, line 35: Is it possible and reasonable to avoid this file-wide `open`?

src/ppx/bucklescript/ppx.ml
```diff
@@ -1,10 +1 @@
...
-  let argc = Array.length Sys.argv in
-  let argv =
-    let output_file = Sys.argv.(argc - 1) in
-    Array.(append
-      (sub Sys.argv 0 (argc - 1))
-      [|"-o"; output_file; "--dump-ast"; "--conditional"|])
-  in
-
-  Migrate_parsetree.Driver.run_main ~argv ()
+let () = Ppxlib.Driver.run_as_ppx_rewriter ()
```
aantron, line 1: This appears to lose the effect of the `--conditional` flag. In your local testing with BuckleScript, are you able to compile the same project xwithout `BISECT_ENABLE=yes` such that instrumentation is disabled, and with `BISECT_ENABLE=yes` such that instrumentation is enabled?

src/ppx/instrument.ml
pitag-ha: That should be possible. But I've just tried and I'm not sure to which extend it would be "reasonable": if we don't open Ppxlib file-wide, OCaml would have a hard time inferring types correctly. E.g. [here](https://github.com/aantron/bisect_ppx/pull/327/files#diff-4bdd650c927dc010be2484e34719df1b48a1c6debc9f8bb81e97f822ae3d7de6R149) it would infer that `e` has type `Parsetree.expression` whereas we want it to be of type `Ppxlib.Parsetree.expression`. And similar in a lot of places. Setting `module Parsetree = Ppxlib.Parstree` wouldn't help in that regard. So apart from overwriting several modules, we'd have to specify a lot of types throughout the whole file.

So in summary, I think it would be possible but quite some work. Do you think it's necessary?

src/ppx/bucklescript/ppx.ml
pitag-ha: The bucklescript test always yields `Coverage: 1/1 (100.00%)`, both with and without `BISECT_ENABLE=yes` and both on my branch and on master. I think the reason for that is that the `conditional` reference is looked at when registering the PPX. At that point it always contains `false`. On master, it gets set to `true` when executing the driver, but at that point the PPX is already registered. But probably I'm missing something. Does using the `--conditional` flag usually have an impact? 

I could set the conditional reference to true here (`Bisect_ppx.Register.conditional := true`), but I have the problem that I'm not sure if that would make a difference. Do you think it would? Do you want me to do that?

Why are there two things in the context of enabling or disabling bisect (the `-conditional` flag and the `BISECT_ENABLE` environment variable)? Does dune set any of them? Or ocamlbuild?

aantron: > The bucklescript test always yields `Coverage: 1/1 (100.00%)`, both with and without `BISECT_ENABLE=yes` and both on my branch and on master.

I can't reproduce that on `master`. Removing `BISECT_ENABLE=yes` from `test/bucklescript/package.json` and doing a clean run of the tests, so that there are no Bisect artifacts left from previous runs with `BISECT_ENABLE=yes`, I get:

```
./node_modules/.bin/bisect-ppx-report html
Error: no coverage files given on command line or found
```



> Does using the `--conditional` flag usually have an impact?
> 
> I could set the conditional reference to true here (`Bisect_ppx.Register.conditional := true`), but I have the problem that I'm not sure if that would make a difference. Do you think it would? Do you want me to do that?

Yes, it has an impact; yes it would make a difference. As for what to do, I originally found it convenient to pass the argument the same way a user would have before the recent improvement to the Dune integration, mainly to ensure I trigger exactly the same logic. I still don't know yet if ppxlib treats arguments the same way. If so, then I suggest passing the argument. If not, you can try setting the reference directly, or using any other reasonable way.

> Why are there two things in the context of enabling or disabling bisect (the `-conditional` flag and the `BISECT_ENABLE` environment variable)? Does dune set any of them? Or ocamlbuild?

For backward compatibility. Originally, Bisect ran unconditionally if it was included in the PPX list at all, and this was fine because it was easy to toggle it from Ocamlbuild. Once Dune came out, we added `--conditional` to make Bisect off by default, since there was no easy way to remove it from the preprocessor list with Dune (short of writing `dune` files in OCaml syntax).

aantron:
> * In my last commit, I changed the version of ocamlformat from 0.15.0 to 0.15.1 to have it compatible with omp2. But I've just noticed that you're using a binary from your fork (so that won't work). Would it be possible to add ocamlformat 0.15.1 or 0.16.0 to your fork to have omp2 compatibility?

Bisect never used 0.15.0, but a specific commit of OCamlFormat, as listed in its dependencies:

https://github.com/aantron/bisect_ppx/blob/2fc5b795e603cde966a45343ef1b523234ce93d4/bisect_ppx.opam#L29-L36

This commit is closer to 0.16.0 than to 0.15.0 or 0.15.1. Looking in the OCamlFormat repo, it was the latest commit prior to the (then-unreleased) 0.16.0 before the switch to omp2. So, you should switch to 0.16.0 instead of 0.15.1. 0.15.1 is behavior-identical with 0.15.0, and is definitely incompatible with 0.16.0 for Bisect's purposes. 0.15.0 is buggy with Bisect's tests &mdash; that's why I had the need to use a git commit in the first place, and AFAIK 0.15.1 preserves the same bugs, just over omp2.

As for the binary, it is a build of the same git commit. I used a misleading tag name for the binary for hosting it on GitHub (I just chose what was the most recent tag in the OCamlFormat repo at the time; looking back, I should have probably given it part of the OCamlFormat commit hash). In any case, you shouldn't have to change the binary, as it should behave the same as 0.16.0 already, and its omp2/non-omp2 dependencies don't matter, as it's pre-built.

You should be able to just change `dune` to have `"ocamlformat" {with-test & = "0.16.0"}` and remove `pin-depends`. I did this locally in testing, and it seems to work.

> * There's one test that's failing locally at the moment (`test/instrument/class/`). It's also failing in master.

I can't reproduce that &mdash; you may be seeing this if you are using OCamlFormat 0.15.1, rather than either the git commit that is listed in Bisect's `dune` and would have been installed by `opam install --deps-only .`, or 0.16.0. If changing to these versions doesn't fix the problem in `master`, could you show the message you get, and confirm the exact OCamlFormat version and/or git commit you have installed when receiving it?

aantron:
> And a third problem: the bucklescript test installation fails on the CI, so I can't see if the bucklescript test would pass or not (locally it did). Do you happen to know what fails in the installation?

The BuckleScript issue looked like a caching/upstream conflict, so I cleared the cache some days ago, and your most recent commit has already passed the BuckleScript tests :)

The opam tests look like they will be fixed, or at least advanced, by updating the OCamlFormat dependency in `dune` to 0.16.0, and by restoring `.travis.yml` to download the existing pre-built binary (as suggested above).

src/ppx/instrument.ml
aantron: I think it's fine to leave it how it is.

aantron: Actually, this is inaccurate:

> it would be possible but quite some work

I just tried to do it myself and got a first version working in less than 2 minutes.

> E.g. [here](https://github.com/aantron/bisect_ppx/pull/327/files#diff-4bdd650c927dc010be2484e34719df1b48a1c6debc9f8bb81e97f822ae3d7de6R149) it would infer that `e` has type `Parsetree.expression` whereas we want it to be of type `Ppxlib.Parsetree.expression`. And similar in a lot of places. Setting `module Parsetree = Ppxlib.Parstree` wouldn't help in that regard. So apart from overwriting several modules, we'd have to specify a lot of types throughout the whole file.

I did exactly what you suggested and it worked. I added

```ocaml
module Parsetree = Ppxlib.Parsetree
module Ast_builder = Ppxlib.Ast_builder
```

and places where metaquot was used needed

```
let open Parsetree in
```

before them.

Some uses of `Nolabel` (which this PR un-qualified) needed to be re-qualified. Overall it was extremely easy to rush through it and get it building. Could you please quickly do a proper version of it?

binaries.esy.json
```diff
@@ -3,12 +3,11 @@
...
-    "@opam/ppx_tools_versioned": "^5.4.0"
+    "@opam/ppxlib": ">= 0.21.0"
   },
   "esy": {
     "build": "dune build -p bisect_ppx",
     "install": "esy-installer bisect_ppx.install",
     "buildsInSource": "_build"
   }
-}
+}
```
aantron, line 13: Please restore this newline.

test/usage/dune
```diff
@@ -0,0 +1 @@
+(dirs (:standard \ reason))
```
aantron, line 1: This file is no longer necessary with recent Reason. At least, I didn't need it locally, and it interferes with running the usage tests (which can be worked around, but better remove this file if possible).

aantron (changes requested):
Overall, the PR looks just about ready, except for the changes from the comments above. So, I added a couple last nits :P

bisect_ppx.opam
```diff
@@ -23,8 +23,7 @@ depends: [
   "cmdliner" {>= "1.0.0"}
   "dune" {>= "2.7.0"}
   "ocaml" {>= "4.02.0"}
-  "ocaml-migrate-parsetree" {>= "1.7.0" & < "2.0.0"}
-  "ppx_tools_versioned" {>= "5.4.0"}
+  "ppxlib" {>= "0.21.0"}
```
aantron, line 26: What is the minimum version with which the non-BuckleScript PPX can be made to work? Is it only the registration interface that the non-bsb code is using right now? Can it be downgraded to loosen this ppxlib constraint?

src/ppx/instrument.ml
pitag-ha: Opening `Parsetree` locally before metaquot uses (instead of type specifications) is a good idea, thanks! And you were right: it has been far less work than I thought.

I've opened `Ocaml_shadow` in the ppx dune file now to make sure that people don't accidentally use compiler-libs in the future. Opening Ppxlib would have given the same protection.

src/ppx/bucklescript/ppx.ml
pitag-ha: Ok, thanks for the explanation! I've set the conditional reference now. Could you maybe check if that does what we want? I've tried and it keeps on doing the same thing for me, but since for me that's also the case on master I must be doing something wrong running the bucklescript test.

At the moment, in Ppxlib it isn't possible to pass arguments to the function that creates the driver. And if it's not necessary, we'd like to avoid that. So I'm trying to find out if it's necessary.

binaries.esy.json
pitag-ha: Good catch

test/usage/dune
pitag-ha: Nice, I didn't know Reason was compatible with OMP2 already.

bisect_ppx.opam
pitag-ha: The code is [accessing the input name through Ppxlib](https://github.com/aantron/bisect_ppx/pull/327/files/6ad06d712cf9f3c96c3638910f0b35a314b1de48#diff-4bdd650c927dc010be2484e34719df1b48a1c6debc9f8bb81e97f822ae3d7de6R1585) for which points 4 and 5 of the [Ppxlib 0.21.0 release notes](https://github.com/ocaml-ppx/ppxlib/releases/tag/0.21.0) are needed. How important is it to loosen the ppxlib constraint?

pitag-ha:
> If changing to these versions doesn't fix the problem in master, could you show the message you get, and confirm the exact OCamlFormat version and/or git commit you have installed when receiving it?

I have Ocamlformat 0.16.0 installed, but the test that's failing doesn't seem to be related to ocamlformat. I'm getting the following on master:
```
+++ b/test/instrument/class/class.t.corrected
@@ -56,9 +56,9 @@ Nested expressions and initializers instrumented.
   >   end
   > EOF
   class foo =
-    let () = ___bisect_post_visit___ 2 (print_endline "bar") in
+    let () = ___bisect_post_visit___ 0 (print_endline "bar") in
     object
       initializer
-      ___bisect_visit___ 1;
-      ___bisect_post_visit___ 0 (print_endline "baz")
+      ___bisect_visit___ 2;
+      ___bisect_post_visit___ 1 (print_endline "baz")
     end
```

bisect_ppx.opam
aantron: Ah, thanks for reminding me of that. It's more important to have access to the input name than to loosen the constraint.

aantron:
> I'll try to figure out how I've broken that test.

This doesn't look like a real failure, but just a different order of traversal used by ppxlib. You should be able to just do `make promote`, commit it, and continue.

src/ppx/bucklescript/ppx.ml
aantron: Can you state exactly what you tried?

For reference:

1. Remove `BISECT_ENABLE=yes` from `package.json` in `test/bucklescript`.
2. Either run `make clean` in `test/bucklescript` (better) or do `rm -f *.coverage` in that directory.

Re-run the BuckleScript test.

src/ppx/instrument.ml
aantron: > I've opened `Ocaml_shadow` in the ppx dune file now to make sure that people don't accidentally use compiler-libs in the future.

Nice, thanks.

aantron:
Looking ahead, you will probably need to remove the 4.02.3 and 4.03.0 rows in Travis. Then, you should take 4.04.2 out of `allow_failures` so that it becomes the new lowest version in the required-to-suceed builds. There are also various mentions of 4.02.3 and 4.03.0 throughout `.travis.yml` that can be cleaned up at that point.

src/ppx/bucklescript/ppx.ml
pitag-ha: What I'm doing is the following:

1. In `src/npm-install.sh`, I've prefixed the path to the driver (lines 8 and 10) with `../../..`. That is because -due to something I'm doing wrong, but I don't know what- the script is in `test/bucklescript/package` at that point and there it can't find the binary. This might be the problem.
2. I've removed `BISECT_ENABLE=yes` from package.json in `test/bucklescript`.
3. From the root of my project I run `make -C test/bucklescript/ full-test` (which should include `make clean`).

pitag-ha: But the way I'm running the test it does work to some extend, since it used to run but fail before fixing Ppxlib and now it passes.

src/ppx/register.ml
```diff
@@ -74,23 +74,22 @@ let switches = [
...
+  |> List.iter (fun (key, spec, doc) -> Ppxlib.Driver.add_arg key spec ~doc)
 
 
 let () =
-  Migrate_parsetree.Driver.register
-    ~name:"bisect_ppx" ~args:switches ~position:100
-    Migrate_parsetree.Versions.ocaml_411 begin fun _config _cookies ->
-      match enabled () with
+  let impl =
+    match enabled () with
```
aantron, line 88: BuckleScript instrumentation is still done unconditionally, unlike in `master`. This is probably because you have moved this `enabled ()` check outside of the suspension (line 90 in the old code) and out here, so that it runs immediately upon module initialization.

This is probably the same reason why the `dune-conditional` integration test is failing here: https://travis-ci.org/github/aantron/bisect_ppx/jobs/757439779#L611-L615. To run these yourself, run `make usage` in the project root directory.

This seems to be the last issue with the PR. I'll take care of the remaining two failures in Travis later. One is about updating some diffs for the self-instrumentation test, and the other looks like a difference (a bug fix) in `-dsource` output causing a spurious test failure on 4.12 (IIRC I reported the original output bug while writing these Bisect tests in the first place, causing harm to Bisect maintenance when it was fixed :P).

src/ppx/bucklescript/ppx.ml
aantron: I've continued the discussion [here](https://github.com/aantron/bisect_ppx/pull/327#pullrequestreview-583006193), since that line looks like the source of the underlying problem with `--conditional` being meaningless in the new code, and there is also a way to test for this using Dune rather than BuckleScript.

src/ppx/register.ml
pitag-ha: Nice! That's what I was meaning before, but I didn't notice that on master that problem didn't exist since there `enabled ()` gets checked inside the callback. I've done the same here now. Thanks! Locally, `make usage` passes now, so I hope on the CI it will, too.

aantron (approved):

aantron:
Thanks @pitag-ha for your work here, and upstream in ppxlib!

PR #364, opened 2021-02-04
Return the AST directly when bisect is disabled

pitag-ha:
It is more efficient to directly return the AST instead of recursing through it without modifying it. I introduced that inefficiency in the port to ppxlib and would like to fix it with this PR.


aantron:
Thanks!

PR #389, opened 2021-11-05
Report: merge coverage files

### LINKED ISSUE #388: Utility to merge coverage files
In our CI we upload several coverage files as artifacts (to create a unified report at the end). However, there is a __lot__ of coverage files and it's too heavy for the CI.

An utility like `bisect-ppx-report merge` could merge every `.coverage` file?

If you agree with this feature, I'd be happy to implement this.

vch9:
Close #388 

I started doing merge by hand but `Input.load_coverage` already does the merge between coverage files.

(I created a dedicated switch, it should not have compatibility issues.)

### DIFF
--- a/src/common/bisect_common.mli
+++ b/src/common/bisect_common.mli
@@ -65,6 +65,9 @@ val register_file :
 
 (** {1 [.coverage] output} *)
 
+val write_runtime_coverage : coverage -> out_channel -> unit
+(** Writes the [coverage] to the given output channel. *)
+
 val write_runtime_data : out_channel -> unit
 (** Writes the current accumulated coverage data (of type {!coverage}) to the
     given output channel. *)
--- a/src/report/main.ml
+++ b/src/report/main.ml
@@ -256,6 +256,16 @@ let coveralls =
 
 
 
+let merge =
+  let call_with_labels
+      to_file coverage_files coverage_paths expect do_not_expect =
+    Merge.output
+      ~to_file ~coverage_files ~coverage_paths ~expect ~do_not_expect
+  in
+  Term.(const set_verbose $ verbose $ const call_with_labels $ to_file
+    $ coverage_files 1 $ coverage_paths $ expect $ do_not_expect),
+  term_info "merge" ~doc:"Merge coverage files"
+
 (* Entry point. *)
 
 let () =
@@ -273,5 +283,5 @@ let () =
           ("See bisect-ppx-report $(i,COMMAND) --help for further " ^
           "information on each command, including options.")
       ]))
-    [html; send_to; text; cobertura; coveralls]
+    [html; send_to; text; cobertura; coveralls; merge]
   |> Term.exit
--- a/src/report/main.ml
+++ b/src/report/main.ml
@@ -257,13 +257,11 @@ let coveralls =
 
 
 let merge =
-  let call_with_labels
-      to_file coverage_files coverage_paths expect do_not_expect =
-    Merge.output
-      ~to_file ~coverage_files ~coverage_paths ~expect ~do_not_expect
+  let call_with_labels to_file coverage_files coverage_paths =
+    Merge.output ~to_file ~coverage_files ~coverage_paths
   in
   Term.(const set_verbose $ verbose $ const call_with_labels $ to_file
-    $ coverage_files 1 $ coverage_paths $ expect $ do_not_expect),
+    $ coverage_files 1 $ coverage_paths),
   term_info "merge" ~doc:"Merge coverage files"
 
 (* Entry point. *)
--- a/src/report/merge.ml
+++ b/src/report/merge.ml
@@ -3,10 +3,10 @@
    https://github.com/aantron/bisect_ppx/blob/master/LICENSE.md. *)
 
 
-let output ~to_file ~coverage_files ~coverage_paths ~expect ~do_not_expect =
+let output ~to_file ~coverage_files ~coverage_paths =
   let coverage =
     Input.load_coverage
-      ~coverage_files ~coverage_paths ~expect ~do_not_expect in
+      ~coverage_files ~coverage_paths ~expect:[] ~do_not_expect:[] in
   let () = Util.mkdirs (Filename.dirname to_file) in
   let oc = open_out to_file in
   try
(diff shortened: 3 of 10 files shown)


src/common/bisect_common.mli
```diff
@@ -65,6 +65,9 @@ val register_file :
 
 (** {1 [.coverage] output} *)
 
+val write_runtime_coverage : coverage -> out_channel -> unit
+(** Writes the [coverage] to the given output channel. *)
+
```
vch9, line 70: This is not ideal. It expose a function only used in `src/report/merge.ml` but it was way easier to call inner function as we have a `coverage`. I believe the other solution would be to build _again_ the coverage using `Bisect_common.register_file`, but this seem too complicated for what we're trying to achieve here.

src/report/merge.mli
```diff
@@ -0,0 +1,16 @@
+(* This file is part of Bisect_ppx, released under the MIT license. See
+   LICENSE.md for details, or visit
+   https://github.com/aantron/bisect_ppx/blob/master/LICENSE.md. *)
+
+
+
+(** This module merge the existing coverage files into a single coverage file *)
```
arvidj, line 7:
```suggestion
(** This module merges coverage files into a single coverage file, summing counters pointwise *)
```

test/report/merge.t
```diff
@@ -0,0 +1,13 @@
...
+
+  $ echo "(lang dune 2.7)" > dune-project
+  $ cat > dune <<'EOF'
+  > (executables
+  >  (names test_merge1 test_merge2)
+  >  (instrumentation (backend bisect_ppx)))
+  > EOF
+  $ dune exec ./test_merge1.exe --instrument-with bisect_ppx
+  $ dune exec ./test_merge2.exe --instrument-with bisect_ppx
+  $ bisect-ppx-report merge merged.coverage
```
arvidj, line 11: you could also test running `bixect-ppx-report summary --per-file` (I think this would run the report by merging the files internally in bisect-ppx-report) and then running `bixect-ppx-report summary --per-file merged.coverage` and verifying that the result is indeed the same.

arvidj (commented):
LGMT. I tried the equivalent of what the cram test does manually and it seems to work fine.

test/report/merge.t
vch9: That's a very good idea! I removed the `cat merged.coverage` as `bisect-ppx-report summary --per-file` is way more readable :)

src/report/main.ml
```diff
@@ -256,6 +256,16 @@ let coveralls =
...
 
 
+let merge =
+  let call_with_labels
+      to_file coverage_files coverage_paths expect do_not_expect =
+    Merge.output
+      ~to_file ~coverage_files ~coverage_paths ~expect ~do_not_expect
+  in
+  Term.(const set_verbose $ verbose $ const call_with_labels $ to_file
+    $ coverage_files 1 $ coverage_paths $ expect $ do_not_expect),
```
aantron, line 266: Do you have a need for `--expect` and `--do-not-expect` in `bisect-ppx-report merge`? I suspect these arguments are only useful when generating a "true" final report. If that's right, I suggest to drop them from the command. You can provide `[]` where `~expect` and `~do_not_expect` are needed internally.

src/report/merge.ml
```diff
@@ -0,0 +1,17 @@
+(* This file is part of Bisect_ppx, released under the MIT license. See
+   LICENSE.md for details, or visit
+   https://github.com/aantron/bisect_ppx/blob/master/LICENSE.md. *)
+
+
+let output ~to_file ~coverage_files ~coverage_paths ~expect ~do_not_expect =
+  let coverage =
+    Input.load_coverage
+      ~coverage_files ~coverage_paths ~expect ~do_not_expect in
```
aantron, line 9: i.e. call this with `~expect:[] ~do_not_expect:[]` if you agree that `--expect` and `--do-not-expect` are not needed for `merge`.

src/report/merge.mli
```diff
@@ -0,0 +1,16 @@
...
+(** This module merges coverage files into a single coverage file, summing counters pointwise *)
+
+
+val output :
+     to_file:string
+  -> coverage_files:string list
+  -> coverage_paths:string list
+  -> expect:string list
+  -> do_not_expect:string list
+  -> unit
```
aantron, line 16: Could you format this in a way that is consistent with the other reporters?

Bisect_ppx uses a style in which:

1. The presence of `->` near arguments is consistent (`->` by every argument, whether it is first or not &mdash; this allows reordering of arguments to be first without causing a spurious diff on the former first argument's line).
2. The return type is visually distinguished from the arguments by indentation and the lack of `->`.
3. The visual left margin is consistent (no big space on one line followed by `->` on another).

https://github.com/aantron/bisect_ppx/blob/8a10004e3b788ff3f681f552f5c6cdbdc3e8f902/src/report/text.mli#L7-L13

test/report/merge.t
```diff
@@ -0,0 +1,26 @@
...
+  > EOF
+  $ dune exec ./test_merge1.exe --instrument-with bisect_ppx
+  $ bisect-ppx-report summary --per-file
+   16.67 %   1/6   merge.ml
+  100.00 %   2/2   test_merge1.ml
+   37.50 %   3/8   Project coverage
+  $ dune exec ./test_merge2.exe --instrument-with bisect_ppx
+  $ bisect-ppx-report summary --per-file
+   33.33 %   2/6    merge.ml
+  100.00 %   2/2    test_merge1.ml
```
aantron, line 17: This

```
   33.33 %   2/6    merge.ml
  100.00 %   2/2    test_merge1.ml
  100.00 %   2/2    test_merge2.ml
```

Looks wrong. I would expect

```
   16.67 %   1/6   merge.ml
  100.00 %   2/2   test_merge2.ml
```

It's probably because the second `bisect-ppx-report summary` is picking up the `.coverage` file from the first `dune exec`, and already implicitly merging them into the report, as it does.

aantron: Ah sorry, I just re-read @arvidj's comment and understood that this is intentional :) Please disregard my comment.

src/report/main.ml
vch9: I do not need them, I removed them in a287e8d7819c23b4b5a3e581f5be3929e005e805

aantron:
Thanks @vch9, and @arvidj for review!

vch9:
> Thanks @vch9, and @arvidj for review!

Thanks again for the quick review :)

Issue #384, opened 2021-10-19
Gracefully handling termination in instrumented binaries


arvidj:
In our system testing framework, we launch processes instrumented with bisect. Normally, these processes terminate normally, writing coverage data as expected. However, some tests may leave unterminated processes, at which point the framework terminates all dangling processes through `SIGTERM`. Then, one of two things may happen:

 - if the unterminated process has not yet started terminating (and if it has no SIGTERM handler of its own eventually calling `exit`), the process will terminate without writing coverage data 
 - if the unterminated process has started terminating, triggering bisect's termination hook, we're seeing corrupted coverage traces (either completely empty, or containing corrupt data).

In the first case we're getting an incomplete coverage measure (as the unterminated process has indeen been tested even though not cleanly terminated). In second, we have the same situation but in addition, `bisect-ppx-report` throws exceptions. 

I don't see any simple solution to this problem (except making sure processes are terminated more gracefully in our systems testing framework), so consider this more of a discussion issue. It seems relatively simple to add a signal handler dealing with the first case, but as you note in https://github.com/aantron/bisect_ppx/issues/122:

> My guess would be that Bisect can't make decisions about threads or signal handlers inside its own runtime. That's too likely to interfere with applications.

For the second case, perhaps the `at_exit` hook can be made more atomic?

arvidj:
In one of our tests touched by this issue, I've worked around the problem by handling signals explicitly in the instrumented process: https://gitlab.com/tezos/tezos/-/merge_requests/3792/

I think this is the way to go for tests that launch their own processes where the code of the child processes is "part of the test itself". 
For tests that launch separate binaries, I think one should:
 - take care to let those binaries terminate gracefully 
 - ensure that those binaries already have signal handling in place that triggers `at_exit` hook on termination 

I agree with you assessment that it would be messy to have bisect mix with the instrumented process's signal handling by default. Nonetheless, I'm afraid we'll have to add such a thing due to potentially large number of binaries we have in our project that do not handle signals at all (or improperly) (the second point above). We might not be able to go through and correct them on a case-by-case basis. Instead, it'd be good to have a blanket solution in the form of the `--mask` and `--sigterm`  options you evoke above. I've thought a bit on how to implement those options in bisect [here](https://gitlab.com/tezos/tezos/-/issues/2002). I'd be happy for your thoughts before I attempt an implementation. 

[...]

PR #390, opened 2021-11-16
Runtime: --bisect-sigterm argument (#384)

arvidj:
Here's a first stab at implementing the `--bisect-sigterm` option as discussed in (#384).

### DIFF
--- a/src/ppx/instrument.ml
+++ b/src/ppx/instrument.ml
@@ -117,6 +117,8 @@ let bisect_file = ref None
 
 let bisect_silent = ref None
 
+let bisect_sigterm = ref false
+
 module Generated_code :
 sig
   type points
@@ -909,6 +911,10 @@ struct
     in
     let bisect_file = ast_convenience_str_opt !bisect_file in
     let bisect_silent = ast_convenience_str_opt !bisect_silent in
+    let bisect_sigterm =
+      let open Parsetree in
+      if !bisect_sigterm then [%expr true] else [%expr false]
+    in
 
     (* ___bisect_visit___ is a function with a reference to a point count array.
        It is called every time a point is visited.
@@ -988,7 +994,7 @@ struct
             let `Visit visit =
               Bisect.Runtime.register_file
                 ~bisect_file:[%e bisect_file] ~bisect_silent:[%e bisect_silent]
-                ~filename:[%e filename] ~points
+                ~filename:[%e filename] ~points ~bisect_sigterm:[%e bisect_sigterm]
             in
             visit
         ]
--- a/src/runtime/native/runtime.ml
+++ b/src/runtime/native/runtime.ml
@@ -132,14 +132,16 @@ let sigterm_handler (_ : int) =
   exit 0
 
 let dump_at_exit () =
-  if not !bisect_file_written then
-    (if !sigterm_enable then
-      (ignore @@ Sys.(signal sigterm Signal_ignore);
-       bisect_file_written := true;
-       dump ();
-       ignore @@ Sys.(signal sigterm Signal_default))
-     else
-       dump ())
+  if not !bisect_file_written then begin
+    if !sigterm_enable then begin
+      ignore @@ Sys.(signal sigterm Signal_ignore);
+      bisect_file_written := true;
+      dump ();
+      ignore @@ Sys.(signal sigterm Signal_default)
+    end
+    else
+      dump ()
+  end
 
 let register_dump : unit Lazy.t =
   lazy (at_exit dump_at_exit)
--- a/src/runtime/native/runtime.ml
+++ b/src/runtime/native/runtime.ml
@@ -10,6 +10,10 @@ let default_bisect_file = ref "bisect"
 
 let default_bisect_silent = ref "bisect.log"
 
+let sigterm_enable = ref false
+
+let bisect_file_written = ref false
+
 type message =
   | Unable_to_create_file
   | Unable_to_write_file
@@ -30,6 +34,14 @@ let full_path fname =
     fname
 
 let env_to_fname env default = try Sys.getenv env with Not_found -> !default
+
+let env_to_boolean ?(default = false) env =
+  try
+    match Sys.getenv env with
+    | "true" -> true
+    | "false" -> false
+    | _ -> default
+  with Not_found -> default
 
 let verbose =
   lazy begin
@@ -114,11 +126,31 @@ let dump () =
         verbose Unable_to_write_file);
       close_out_noerr channel
 
-let register_dump : unit Lazy.t =
-  lazy (at_exit dump)
+let sigterm_handler (_ : int) =
+  bisect_file_written := true;
+  dump ();
+  exit 0
 
-let register_file ~bisect_file ~bisect_silent ~filename ~points =
+let dump_at_exit () =
+  if not !bisect_file_written then
+    (if !sigterm_enable then
+      (ignore @@ Sys.(signal sigterm Signal_ignore);
+       bisect_file_written := true;
+       dump ();
+       ignore @@ Sys.(signal sigterm Signal_default))
+     else
+       dump ())
+
+let register_dump : unit Lazy.t =
+  lazy (at_exit dump_at_exit)
+
+let register_sigterm_hander : unit Lazy.t =
+  lazy (ignore @@ Sys.(signal sigterm (Signal_handle sigterm_handler)))
+
+let register_file ~bisect_file ~bisect_silent ~bisect_sigterm ~filename ~points =
   (match bisect_file with None -> () | Some v -> default_bisect_file := v);
   (match bisect_silent with None -> () | Some v -> default_bisect_silent := v);
+  sigterm_enable := env_to_boolean ~default:bisect_sigterm "BISECT_SIGTERM";
+  (if !sigterm_enable then Lazy.force register_sigterm_hander);
   let () = Lazy.force register_dump in
   Common.register_file ~filename ~points
--- a/src/runtime/native/runtime.mli
+++ b/src/runtime/native/runtime.mli
@@ -29,6 +29,10 @@
     routed to [stderr], otherwise [BISECT_SILENT] is used to determine a
     filename for this output. The default value is [bisect.log].
 
+    If the environment variable [BISECT_SIGTERM] is set to [true], then
+    the runtime will install a signal handler for [SIGTERM], writing
+    coverage output before terminating.
+
     Because instrumented modules refer to [Bisect], one is advised to link
     this module as one of the first ones of the program.
 
@@ -41,16 +45,18 @@
 val register_file :
   bisect_file:string option ->
   bisect_silent:string option ->
+  bisect_sigterm:bool ->
   filename:string ->
   points:int list ->
     [`Visit of (int -> unit)]
-(** [register_file ~bisect_file ~bisect_silent file ~point_count
+(** [register_file ~bisect_file ~bisect_silent ~bisect_sigterm file ~point_count
     ~point_definitions] indicates that the file [file] is part of the
     application that has been instrumented. [point_definitions] is a serialized
     [Common.point_definition list] giving the locations of all points in the
     file. The returned callback is used to increment visitation counts.
-    [bisect_file] (resp. [bisect_silent]) is a default value for the environment
-    variable [BISECT_FILE] (resp. [BISECT_SILENT]). *)
+    [bisect_file] (resp. [bisect_silent] and [bisect_sigterm]) is a default
+    value for the environment variable [BISECT_FILE] (resp. [BISECT_SIGTERM] and
+    [BISECT_SIGTERM]). *)
 
 val get_coverage_data : unit -> string option
 (** Returns the binary coverage data accumulated by the program so far. This
--- a/src/ppx/instrument.mli
+++ b/src/ppx/instrument.mli
@@ -20,3 +20,6 @@ val bisect_file : string option ref
 
 val bisect_silent : string option ref
 (** Default value for [BISECT_SILENT]. *)
+
+val bisect_sigterm : bool ref
+(** Default value for [BISECT_SIGTERM]. *)
--- a/src/ppx/register.ml
+++ b/src/ppx/register.ml
@@ -44,6 +44,11 @@ let switches = [
   ("--bisect-silent",
   Arg.String (fun s -> Instrument.bisect_silent := Some s),
   " Default value for BISECT_SILENT environment variable");
+
+  ("--bisect-sigterm",
+  Arg.Set Instrument.bisect_sigterm,
+  (" Install a signal handler writing coverage data and" ^
+   " terminating on reception of SIGTERM"));
 ]
 
 let () =
--- a/src/runtime/bucklescript/runtime.ml
+++ b/src/runtime/bucklescript/runtime.ml
@@ -43,6 +43,7 @@ let write_coverage_data_on_exit () =
     exit_hook_added := true
   end
 
-let register_file ~bisect_file:_ ~bisect_silent:_ ~filename ~points =
+let register_file
+    ~bisect_file:_ ~bisect_silent:_ ~bisect_sigterm:_ ~filename ~points =
   write_coverage_data_on_exit ();
   Bisect_common.register_file ~filename ~points
--- a/src/runtime/bucklescript/runtime.mli
+++ b/src/runtime/bucklescript/runtime.mli
@@ -7,6 +7,7 @@
 val register_file :
   bisect_file:string option ->
   bisect_silent:string option ->
+  bisect_sigterm:bool ->
   filename:string ->
   points:int list ->
     [`Visit of (int -> unit)]
@@ -16,7 +17,7 @@ val register_file :
     [Common.point_definition list] giving the locations of all points in the
     file. The returned callback is used to increment visitation counts.
 
-    [~bisect_file] and [~bisect_silent] are ignored. *)
+    [~bisect_file], [~bisect_silent], and [~bisect_sigterm] are ignored. *)
 
 val get_coverage_data : unit -> string option
 (** Returns the binary coverage data accumulated by the program so far. This
--- a/test/instrument/recent/exclusions.t
+++ b/test/instrument/recent/exclusions.t
@@ -42,7 +42,7 @@
         let points = [12] in
         let `Visit visit =
           Bisect.Runtime.register_file ~bisect_file:None ~bisect_silent:None
-            ~filename:"not_excluded.ml" ~points in
+            ~filename:"not_excluded.ml" ~points ~bisect_sigterm:false in
         visit
       let ___bisect_post_visit___ point_index result =
         ___bisect_visit___ point_index; result
--- a/test/instrument/structure.t
+++ b/test/instrument/structure.t
@@ -8,7 +8,7 @@ An empty file. Show the bare registration code.
       let points = [] in
       let (`Visit visit) =
         Bisect.Runtime.register_file ~bisect_file:None ~bisect_silent:None
-          ~filename:"test.ml" ~points
+          ~filename:"test.ml" ~points ~bisect_sigterm:false
       in
       visit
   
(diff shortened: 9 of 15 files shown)


src/runtime/native/runtime.ml
```diff
@@ -10,16 +10,33 @@ let default_bisect_file = ref "bisect"
 
 let default_bisect_silent = ref "bisect.log"
 
+let sigterm_enable = ref false
+
+let sigterm_ignore_overriden_warned = ref false
+
+let sigterm_ignore_handle_warned = ref false
```
arvidj, line 17: I've added these warnings that are emitted if an existing signal handler is overwritten. 
To not output them in each file visited, I track their emission in these references.

src/runtime/native/runtime.ml
```diff
@@ -10,16 +10,33 @@ let default_bisect_file = ref "bisect"
...
+
+let bisect_file_written = ref false
```
arvidj, line 19: We need to track whether the file has already been written. To terminate the application when receiving `SIGTERM`, the signal handler calls `exit`, triggering the `at_exit` hook. To avoid this, we set `bisect_file_written` to true in `dump`, and do nothing when it is set to true.

src/runtime/native/runtime.ml
```diff
@@ -101,24 +124,44 @@ let dump_counters_exn =
...
-  | Some channel ->
-      (try
-        dump_counters_exn channel
-      with _ ->
-        verbose Unable_to_write_file);
-      close_out_noerr channel
+  if not !bisect_file_written then
+    (if !sigterm_enable then
+       ignore @@ Unix.(sigprocmask SIG_BLOCK [Sys.sigterm]) ;
+     bisect_file_written := true;
```
arvidj, line 130: i set it to true regardless whether the write succeeds or not, not sure that makes sense.

src/runtime/native/runtime.ml
```diff
@@ -101,24 +124,44 @@ let dump_counters_exn =
...
 let register_dump : unit Lazy.t =
   lazy (at_exit dump)
 
-let register_file ~bisect_file ~bisect_silent ~filename ~points =
+let register_file ~bisect_file ~bisect_silent ~bisect_sigterm ~filename ~points =
   (match bisect_file with None -> () | Some v -> default_bisect_file := v);
   (match bisect_silent with None -> () | Some v -> default_bisect_silent := v);
+  sigterm_enable := env_to_boolean "BISECT_SIGTERM" bisect_sigterm;
+  (if !sigterm_enable then
+    match Sys.(signal sigterm (Signal_handle sigterm_handler)) with
```
arvidj, line 156: this code is a bit ugly due to the warnings i've added, if they're not necessary then it can be simplified.

src/ppx/instrument.ml
```diff
@@ -909,6 +911,10 @@ struct
     in
     let bisect_file = ast_convenience_str_opt !bisect_file in
     let bisect_silent = ast_convenience_str_opt !bisect_silent in
+    let bisect_sigterm =
+      let bool_s = (if !bisect_sigterm then "true" else "false") in
+      Exp.construct ~loc {txt = Longident.parse bool_s; loc} None
```
vch9, line 916:
```suggestion
      if !bisect_sigterm then [%expr true] else [%expr false]
```

test/usage/dune-sigterm/test-daemon.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -e
+
+dune exec ./daemon.exe &
+# Wait a bit until the application has started
+sleep 1
```
arvidj, line 7: it is a bit unfortunate that i have to do this, but if i do not sleep it seems the application does not have enough time to start up and install the signal handler. putting a lower value for sleep does not seem to work very well. suggestions? i could have `daemon.exe` communicate with the outside world when it has started (write a file or something) and wait for that instead.

src/runtime/native/runtime.ml
```diff
@@ -31,6 +48,12 @@ let full_path fname =
 
 let env_to_fname env default = try Sys.getenv env with Not_found -> !default
 
+let env_to_boolean env default =
```
vch9, line 51:
```suggestion
let env_to_boolean env ?(default = false) =
```

arvidj: `Warning 16: this optional argument cannot be erased.`

vch9: How's that possible? You have `env` as the last argument? (Before editing I did put `default` as the last argument).

test/usage/dune-sigterm/test-daemon.sh
vch9: I think it's ok. Alternatively, would the corrupted file happens if you have an OCaml code that does a `fork` and kill its child?

aantron:
I think it's best to ignore this in the ReScript runtime for now. Also, don't worry about errors in that &mdash; I can adapt it later.

src/runtime/native/runtime.mli
```diff
@@ -41,16 +45,18 @@
...
     [`Visit of (int -> unit)]
-(** [register_file ~bisect_file ~bisect_silent file ~point_count
+(** [register_file ~bisect_file ~bisect_silent ~bisect_sigterm file ~point_count
     ~point_definitions] indicates that the file [file] is part of the
     application that has been instrumented. [point_definitions] is a serialized
     [Common.point_definition list] giving the locations of all points in the
     file. The returned callback is used to increment visitation counts.
-    [bisect_file] (resp. [bisect_silent]) is a default value for the environment
-    variable [BISECT_FILE] (resp. [BISECT_SILENT]). *)
+    [bisect_file] (resp. [bisect_silent] and [bisect_sigterm ) is a default
```
aantron, line 57: The markup `[bisect_sigterm )` looks wrong here.

src/ppx/instrument.ml
aantron: Agree with this, but it also needs `let open Parsetree in` on my machine.

aantron: As here:

https://github.com/aantron/bisect_ppx/blob/c3b8b5cd6f74ecbcfb2b3a4ae2b172e3f1fd93a6/src/ppx/instrument.ml#L168-L175

src/runtime/native/runtime.ml
aantron: I would lean towards omitting these warnings, unless you already foresee that you might have difficulty with silent replacement of SIGTERM handlers in your own usage.

A user that is using `--bisect-sigterm` or `BISECT_SIGTERM` has already explicitly opted into having Bisect install a SIGTERM handler. If we ever have a real user that has a complex enough assortment of binaries in testing that they would benefit from this warning, we can add the warnings then.

It's also likely that rather than warning, we would want to "chain" to the previous signal behavior &mdash; not exit the process if the signal was ignored, or call the other function if there was already a handler.

However, a binary under test that installs its own signal handlers outside the Bisect runtime is likely not to interact well with `--bisect-sigterm` anyway, since it will tend to install its own signal handler *after* `--bisect-sigterm`. That means that both the warnings and attempts to chain signal handlers are likely not to work anyway, since `--bisect-sigterm` will be processed first, and then the binary will clobber `--bisect-sigterm` behavior.

src/runtime/native/runtime.ml
```diff
@@ -101,24 +124,43 @@ let dump_counters_exn =
...
-  | None -> ()
-  | Some channel ->
-      (try
-        dump_counters_exn channel
-      with _ ->
-        verbose Unable_to_write_file);
-      close_out_noerr channel
+  if not !bisect_file_written then
+    (if !sigterm_enable then
+       ignore @@ Unix.(sigprocmask SIG_BLOCK [Sys.sigterm]) ;
```
aantron, line 129: Blocking of the signal should be moved to a wrapper that will be called only during `at_exit`. As this code is now, the signal will be blocked even if `dump` is called from the SIGTERM handler. This is not necessary, as SIGTERM should already be blocked during the SIGTERM handler.

I also suggest to unblock SIGTERM at the end of the new wrapper around `dump`, so that if there are other `at_exit` handlers in the binary under test, they can be interrupted by SIGTERM as normal.

aantron: Also, I think you can stick with `Sys.signal` here, rather than using `Unix.sigprocmask`.

test/usage/dune-sigterm/test-daemon.sh
aantron: `sleep 1` is fine for this test.

test/usage/dune-sigterm/Makefile
```diff
@@ -0,0 +1,23 @@
...
+## TODO: test environment variable as well
+
+
+.PHONY : test
+test : test-normal test-daemon
+
+.PHONY : clean
+clean :
+	dune clean
+	rm -f bisect*.coverage
```
aantron, line 23: Is the choice to create a usage test (intended so far for testing build systems) deliberate, as opposed to creating a cram test?

src/ppx/instrument.ml
arvidj: Fixed in c858d69.

src/runtime/native/runtime.ml
arvidj: I messed up. Here's the fix: a7db78e

src/runtime/native/runtime.ml
arvidj: Makes sense to me. It can be useful when you're instrumenting a big application and you don't actually know whether it installs it's own signal handlers or not. 

Warnings removed in 04c164d.

src/runtime/native/runtime.mli
arvidj: Fixed in 346c9d2.

src/runtime/native/runtime.ml
```diff
@@ -101,24 +113,35 @@ let dump_counters_exn =
...
 
 let register_dump : unit Lazy.t =
   lazy (at_exit dump)
 
-let register_file ~bisect_file ~bisect_silent ~filename ~points =
+let register_file ~bisect_file ~bisect_silent ~bisect_sigterm ~filename ~points =
   (match bisect_file with None -> () | Some v -> default_bisect_file := v);
   (match bisect_silent with None -> () | Some v -> default_bisect_silent := v);
+  sigterm_enable := env_to_boolean ~default:bisect_sigterm "BISECT_SIGTERM";
+  (if !sigterm_enable then
```
aantron, line 144: I believe this will currently install the signal handler each time a source file is registered with the Bisect runtime. It should probably be guarded by some `ref` so that the signal handler is installed only once.

test/usage/dune-sigterm/Makefile
arvidj: nope, that was an oversight from my part. in the b425545 i move it to a cram test. i also changed the sleep thing (it was fickle) and replaced it with the fork suggestion of valentin. i was not sure in which folder to put the test, i've put it in `test/sigterm/` but i can move it somewhere else that makes more sense.

aantron:
The cram test looks good, thanks!

It looks like this PR has only the two remaining issues about [when exactly SIGTERM is blocked](https://github.com/aantron/bisect_ppx/pull/390#discussion_r750507135) and [whether the handler is installed repeatedly](https://github.com/aantron/bisect_ppx/pull/390#discussion_r751102411), which should only need pretty small alterations. After that, is this code working in your test environment?

The ReScript test failures and the difference in spacing of `wc` output on Mac can be ignored. I can fix that later.

src/runtime/native/runtime.ml
arvidj: In 3d413f8 I wrap the registration in a lazy, like is done for the `at_exit` registration. Is that OK?

src/runtime/native/runtime.ml
arvidj: In 0f87a8f I unblock SIGTERM after `at_exit`. I also add a (convoluted) test that verifies that succeeding at_exit hooks can be interrupted.

arvidj:
Yes, latest version seems to be working fine in our test suite.

src/runtime/native/runtime.ml
arvidj: I'm not sure I understand what you mean here. Actually, we need just need to track if `dump` has been called to avoid writing the trace twice. However, I've changed the logic a bit in the latest version, perhaps it is more clear there.

src/runtime/native/runtime.ml
```diff
@@ -114,11 +126,30 @@ let dump () =
         verbose Unable_to_write_file);
       close_out_noerr channel
 
+let sigterm_handler (_ : int) =
+  bisect_file_written := true;
+  dump ();
+  exit 0
+
+let dump_at_exit () =
```
aantron, line 134: Last remaining issue. I think this function is now missing a `bisect_file_written := true`, which should probably come after the masking of the signal in the `!sigterm_enable` case.

aantron: However, maybe I am misunderstanding something, because looking back at the new cram test, I would have expected 3 coverage files, one from the parent and two from the child, but the test shows the "proper" 2 files. Not sure what is going on.

arvidj: makes sense, fixed in the last version

aantron: Ah, I believe we are not observing anything because the last function added with `at_exit` is the one added by the "user" program in the test, so it runs first, and SIGTERM comes during that function. That also means that the test isn't actually testing re-enabling of signals.

aantron: The Bisect runtime's `at_exit` registration comes so early in the initialization of the program (since it is inserted at the start of every instrumented source file) that it would be difficult to make it run not last. I think you would have to do something like have a non-instrumented sub-library that the instrumented binary depends on, and the sub-library would have to do `at_exit`. That seems pretty excessive a burden for a test in a PR. I suggest to remove the current test, and I'll add the fancy early `at_exit` test later on.

aantron: Or we can just leave the test and I'll fix it up later :) Anyway, I'm ready to merge now, let me know what you want to do or mark the PR as non-draft :)

arvidj: Ah, my brain is fried right now so I'll have to check tomorrow. I know at least that if I do not re-enable the signal, then the new cram test fails.

arvidj: > I think you would have to do something like have a non-instrumented sub-library that the instrumented binary depends on, and the sub-library would have to do at_exit.

that's what I've attempted to do, notice that `at_exit_hook.ml` is not instrumented.

aantron: Ah. But AFAICT `at_exit_hook.ml` is still instrumented, because the `dune` file [emitted](https://github.com/aantron/bisect_ppx/pull/390/files#diff-ebcf58b941ad53f583173fdee96923dca6ff3ecbdb1f20ef8c85f13c56169361R19-R21) in the test still treats the whole set of `.ml` files as just an executable and the `preprocess` stanza is applying to all the `.ml` files pulled in during the dependency scan, since they get treated as part of the executable (there is no `library` stanza).

Ok, can wait for tomorrow. My brain is also a bit fried :) However, I also want to note that needing to fiddle a bit with a very edge case test doesn't block merging a PR, IMO :)

src/runtime/native/runtime.ml
aantron: Yes, thank you!

src/runtime/native/runtime.ml
aantron: Well I somehow missed the `[@@@coverage exclude_file]`! Sorry. Ok, at this point I am just confused about that test, which makes me all the more ready to merge this PR so I can do an investigation later, unless you strongly prefer to do it on your end :)

arvidj: Here is the order of how things run, as per a sprinkling of
`print_endline`:

```
+  at_exit_hook.ml: register at_exit (before fork)
+   *** (499322) register sigterm handler
+   *** (499322) register dump
+  at_exit_main.ml: parent: start
+  at_exit_main.ml: child: finished
+   *** (499486) in dump_at_exit
+  at_exit_hook.ml: child sleeps
+  at_exit_main.ml: parent: kill child
+  at_exit_main.ml: parent: finished
+   *** (499322) in dump_at_exit
```

so the top-level definitions of `at_exit_hook.ml` are evaluated first
(as it is referred to by `at_exit_main.ml`). this module does not
contain the bisect instrumentation (as can be verified with `$ ocamlc
-dsource ../../_build/default/test/sigterm/at_exit_hook.pp.ml`).  )

so all it does is register the applications at_exit hook.

then, the initialization of bisect_ppx runs when the top-level
definitions of `at_exit_main.ml` are evaluated: the sigterm handler
and dump at_exit hook are installed.

then the application starts. it forks and the child exits immediately.
the at_exit hooks installed now are `[bisect; application]` and they
are executed in LIFO. hence bisect's hook runs first. then the child's
hook runs. it sleeps.

meanwhile, the parent process kills the child. if the child did not
have it's default sigterm restored, then this kill would be ignored
and we should see the message "I should've been woken up by now.". 

the number of traces written are two: one from the at_exit hook in each process.

does that make sense? i'd be happy to make something simpler if you have any nice ideas.

otherwise, i don't mind if you merge it as it is, and we can iron out any wrinkles later on.

arvidj:
squashed

aantron:
Thank you! I will do a 2.7.0 release soon.

arvidj:
> Thank you! I will do a 2.7.0 release soon.

Cool, thanks!

> squashed

i just meant that i rewrote the history into one commit. Do you prefer to keep it in smaller commits? (for future mrs ;))

aantron:
> i just meant that i rewrote the history into one commit. Do you prefer to keep it in smaller commits? (for future mrs ;))

It's not necessary to squash (at least on GitHub), since GitHub has a merge-by-squashing button (which I use :P). On the other hand, squashing loses the continuity of review, i.e. during review the reviewer builds up an understanding up to the last commit in the PR. If the PR is then squashed (without agreement), the new commit is potentially an all-new diff and it can take work to make sure that the diff in that new commit is the same as the net diff the reviewer has already reviewed. So I usually ask not to force push into PRs after review has begun, except by agreement (it's fine to force push to replace a commit immediately after first pushing it, to quickly fix some errors if needed).

The vast majority of PRs are well addressed by GitHub's built-in squash-merge, turning them into one commit regardless of how messy the PR branch history was, and without the need for rewriting that history inside the PR (and threatening review). A few make more sense to merge as several commits, but then the maintainer can either do that themselves, or discuss with the PR author.

arvidj:
> > i just meant that i rewrote the history into one commit. Do you prefer to keep it in smaller commits? (for future mrs ;))
> 
> It's not necessary to squash (at least on GitHub), since GitHub has a merge-by-squashing button (which I use :P). ...

Ah ok, I see. Sorry about that, I'm not so used to github, on the tezos/tezos repo we usually have authors squash their MRs into a single commit (for smaller stuff) or rewrite their history into a clean set of atomic commits (e.g. no fixups). Noted for future contributions!

aantron:
Yep, this was also the practice on GitHub in the past (before squash-merge, when GitHub also seemed to be philisophically completely against squashing by anyone ever, didn't allow pushing into contributor branches, and it had to be negotiated :P). And in the case of this PR the squash didn't cause any problems, since I already had the branch locally for testing, so I was able to trivially do a `diff` to double-check all is well :)

Thanks again!

aantron:
FYI I've [changed](https://github.com/aantron/bisect_ppx/commit/c606462c3ea34da70e1efa4312ddad5078e18a9b) `BISECT_SIGTERM` to use values `yes`/`no` rather than `true`/`false`, for consistency with Bisect's existing environment variables `BISECT_ENABLE` and `BISECT_SILENT`.

aantron:
This and other recent changes are now in opam in [Bisect_ppx 2.7.0](https://github.com/aantron/bisect_ppx/releases/tag/2.7.0).

arvidj:
Great, thanks a lot Anton!

PR #395, opened 2022-01-19
Points in an array instead of list

jfrolich:
In ReScript lists are compiled as a nested data structure, resulting in a very deep data structure. In large files this causes babel to stack overflow. With this change, the points are saved as an array, which is represented as a simple JavaScript array.

### DIFF
--- a/src/common/bisect_common.ml
+++ b/src/common/bisect_common.ml
@@ -9,7 +9,7 @@
 
 type instrumented_file = {
   filename : string;
-  points : int list;
+  points : int array;
   counts : int array;
 }
 
@@ -37,7 +37,7 @@ let write_list write_element formatter l =
 
 let write_instrumented_file formatter {filename; points; counts} =
   write_string formatter filename;
-  write_list write_int formatter points;
+  write_list write_int formatter (Array.to_list points);
   write_array write_int formatter counts
 
 let write_coverage formatter coverage =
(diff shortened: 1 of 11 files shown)


src/common/bisect_common.ml
```diff
@@ -37,7 +37,7 @@ let write_list write_element formatter l =
 
 let write_instrumented_file formatter {filename; points; counts} =
   write_string formatter filename;
-  write_list write_int formatter points;
+  write_list write_int formatter (Array.to_list points);
```
aantron, line 40: Could you use `write_array` for this instead, as done on the next line?

src/ppx/instrument.ml
```diff
@@ -890,15 +890,9 @@ struct
...
-        let rec inline_point_offsets acc = function
-        | [] -> acc
-        | offset::more ->
-          inline_point_offsets
-            [%expr [%e Ast_builder.Default.eint ~loc offset]::[%e acc]]
-            more
-      in
-      inline_point_offsets [%expr []] points.offsets
+        Ast_builder.Default.pexp_array ~loc (List.map (
+          fun offset -> Ast_builder.Default.eint ~loc: Location.none offset
```
aantron, line 894: The original code used `loc` for `~loc` for the elements. Is there a good reason to replace that with `Location.none`, as done here?

aantron (commented):
Thanks! Have you tested this on your use case, and it solves the Babel stack overflow problem?

src/ppx/instrument.ml
jfrolich: No I thought the elements didn't have a loc, but was mistaken

Issue #334, opened 2020-10-10
Fix js_of_ocaml integration


aantron:
The js_of_ocaml integration test has recently started failing with

```
            throw _k_}}}
            ^
0,248,Failure,-2,TypeError: runtime.unix_open is not a function
```

aantron:
This is caused by the fix of https://github.com/ocaml/dune/issues/3622, which led to js_of_ocaml programs linked by Dune calling their `at_exit` hooks. Bisect_ppx's js_of_ocaml integration was (and as of right now, still is) relying on those hooks not being called, as they contain Unix-specific code.

pat-rondon:
Thanks for making this interoperable with JSOO!

There's an additional tweak that's needed for binaries linked with the JSOO runtime library, and which thus have type [`Jsoo_runtime.Error`](https://ocsigen.org/js_of_ocaml/latest/api/js_of_ocaml-compiler/Jsoo_runtime/Error/index.html) registered, as described in https://github.com/ocsigen/js_of_ocaml/issues/1146. Binaries linked with the runtime library will throw an exception of type `Jsoo_runtime.Error` rather than `Failure`. (But binaries _not_ linked with the runtime library continue to throw `Failure`.)

A test demonstrating the issue and potential fix are in #397.

PR #397, opened 2022-02-06
Condition Runtime.dump execution for JSOO-compiled binaries on the compiler backend used rather than exception raised

pat-rondon:
Generalizes the fix of #334 for js_of_ocaml_integration:
- Adds a test demonstrating that, when running under js_of_ocaml and linking its runtime library, the expected `Failure` exception is not raised, and a `Jsoo_runtime.Error` is raised instead. This test is identical to the existing `js_of_ocaml` test except that it also links in the JSOO runtime library.
- Changes the logic for skipping the call to `Runtime.dump` to skip the call when the backed is `js_of_ocaml`, rather than when the `Failure` exception is raised.


aantron:
Thank you!

I'll wait a bit for #396 and then do a release with this.

pat-rondon:
Thanks!

PR #399, opened 2022-02-17
Add support for OCaml 5.00

kit-ty-kate:


aantron:
Thanks!

CI failures look like an unrelated test that has become flaky.

PR #401, opened 2022-05-16
Fix collation of subdirectories in HTML report

lukemaurer:
We have a project with top-level source files as well as subdirectories, and the listing was being cut off before the first subdirectory. This came down to the `partition_files` function, which did not correctly handle directories containing both files and subdirectories. (It also returned the directories in reverse order, which this patch also fixes.)

### DIFF
--- a/src/report/html.ml
+++ b/src/report/html.ml
@@ -85,25 +85,25 @@ let output_html_index ~tree title theme filename files =
       let directory', _ = split_filename name in
       directory' = directory
     in
-    let rec aux sub_dirs =
+    let rec aux sub_dirs sub_files =
       function
-        [] -> (sub_dirs, [])
+        [] -> (List.rev sub_dirs, List.concat (List.rev sub_files))
       | (name, _, _) as file :: files ->
          match subdirectory_of ~directory name with
          | None ->
-            let (sub_files, _) =
+            let (sub_files_here, files') =
               Util.split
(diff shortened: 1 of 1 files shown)


vch9:
Hi and thanks for the patch!
We've been having the same issue for weeks/months and I tried to investigate a bit. After producing a few debugging traces I tried your patch and it fixed the problem on our side (the cause is the same as yours)

I believe this closes https://github.com/aantron/bisect_ppx/issues/409?

What do you think @aantron?

vch9:
Also note that this impacts the coverage %:

Without this patch on master
```
$ dune exec vendors/bisect_ppx/src/report/main.exe --  summary --coverage-path .                    
Coverage: 60470/285629 (21.17%)

$ dune exec vendors/bisect_ppx/src/report/main.exe --  html --ignore-missing-files --coverage-path .
> 19.10% (taken from the index)
```

With the patch:
```
$ dune exec vendors/bisect_ppx/src/report/main.exe --  summary --coverage-path .                    
Coverage: 60470/285629 (21.17%)

$ dune exec vendors/bisect_ppx/src/report/main.exe --  html --ignore-missing-files --coverage-path .
> 21.17% (taken from the index)
```

aantron:
Thanks @lukemaurer and @vch9! I am currently finishing up bugfixes in another project and will review this shortly!

aantron:
Thank you!

PR #405, opened 2022-08-15
Fix for `rescript 10.0.1`

quinn-dougherty:
Taking a quick stab at this: https://github.com/aantron/bisect-starter-rescript/issues/3 

when you try `rescript` 10 with `bisect_ppx`: 
```
 $ rescript
Dependency on bisect_ppx
rescript: [1/4] src/common/bisect_common-Bisect.cmi
FAILED: src/common/bisect_common-Bisect.cmi
File "/home/.../bisect-starter-rescript/node_modules/bisect_ppx/src/common/bisect_common.mli", line 68, characters 42-53:
Error: Unbound type constructor out_channel
```

I piggy-backed a `flake.nix` for a reproducible development environment, mostly for personal use. We can cherry-pick it out of the PR if we don't want it.


quinn-dougherty:
Confused why `out_channel` can be unbounded, since it's clearly in stdlib which apparently is `open`'d at the top of every file.   https://v2.ocaml.org/releases/4.13/api/Stdlib.html#TYPEout_channel

@christianoc told me on https://github.com/rescript-lang/rescript-compiler/issues/5596 that some things to do with `Pervasives` (which is the same thing as `Stdlib`) were removed from `rescript` 10.0, I also notice in the `rescript-compiler` `CONTRIBUTING.md` file it says that ocaml 4.14 is used, but I don't think that's related.

quinn-dougherty:
The issue: https://github.com/aantron/bisect_ppx/issues/406

cristianoc:
Would this same PR work with the older version of ReScript?
The new version should only affect compiling ReScript projects. What part is currently failing?

dsiu:
Any quick fix for this for ReScript?

cristianoc:
Great! Do you mind testing v11 beta too? It should just work, but just in case.

aantron:
Thanks for the heads up! I've fixed some incompatibilities in https://github.com/aantron/bisect_ppx/commit/9439ba71f1c138ca38229affa180ffef6aa3b978, but I'm currently left with:

```
λ make -C test/js full-test

[...]

> node ./lib/js/hello.js

internal/modules/cjs/loader.js:818
  throw err;
  ^

Error: Cannot find module '/home/antron/code/bisect/bisect_ppx/test/js/lib/js/hello.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:815:15)
    at Function.Module._load (internal/modules/cjs/loader.js:667:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

This is the former location of the output, `hello.js`, with ReScript <= 10. As far as I can tell, there is no output file at all now:

```
λ cd test/js && find lib                                                                                                                                          2

lib
lib/bs
lib/bs/.bsdeps
lib/bs/.bsbuild
lib/bs/.compiler.log
lib/bs/build.ninja
lib/bs/.ninja_log
lib/bs/install.ninja
lib/bs/.sourcedirs.json
lib/js
```

Could you suggest what changed in ReScript 11? I wasn't able to spot a relevant change in the changelogs for the ReScript 11 pre-release versions.

cristianoc:
Would a compiler error (type checking or other) leave some trace here?
One difference could be partial application is not possible by default, but there are quite a few changes, so hard to say without looking at 1 individual compilation.

aantron:
There were several compiler errors that I fixed along the way. As I understood your message, the file should be at the same place, but there is probably some silent error along the way that prevents it from being generated. I'm going to open a separate issue to look into that later. Thank you!

aantron:
The issue was that ReScript no longer does anything with `.re` files, so there were no compilation units detected and thus no output.

Issue #416, opened 2023-05-28
Proposal: add <packages> to cobertura report


maxim092001:
Hey!

## Proposal:

I would like to add `<packages>` around packages.

For example current report generated by bisect_ppx looks like:

```xml
<?xml version="1.0"?>
<coverage lines-valid="282" lines-covered="79"
line-rate="0.280142">
  <sources>
    <source>.</source>
  </sources>
  <package name="." line-rate="0.280142">
    <classes>
      <class name="lib/message.ml" filename="lib/message.ml"
      line-rate="0.250000">
        <lines>
          <line number="4" hits="4" />
          <line number="6" hits="0" />
          <line number="7" hits="0" />
          <line number="8" hits="0" />
        </lines>
      </class>
    </classes>
  </package>
</coverage>
```

And I would like it to be as:

```xml
<?xml version="1.0"?>
<coverage lines-valid="282" lines-covered="79"
line-rate="0.280142">
  <sources>
    <source>.</source>
  </sources>
  <packages>
    <package name="." line-rate="0.280142">
      <classes>
        <class name="lib/message.ml" filename="lib/message.ml"
        line-rate="0.250000">
          <lines>
            <line number="4" hits="4" />
            <line number="6" hits="0" />
            <line number="7" hits="0" />
            <line number="8" hits="0" />
          </lines>
        </class>
      </classes>
    </package>
  </packages>
</coverage>
```

## Why:

Some tools used to generate HTML or any other representation of report require `<packages>`. For example: [pycobertura](https://github.com/aconrad/pycobertura/tree/master)

Also, in [main cobertura repo](https://github.com/cobertura/cobertura/blob/0ff963284cecaace30f409a977dccb07c41a5a8f/cobertura/src/main/java/net/sourceforge/cobertura/reporting/xml/XMLReport.java#L138) they also have `<packages>`. 

Also, `<packages>` is a part of [some kind of official specification](https://github.com/cobertura/cobertura/blob/0ff963284cecaace30f409a977dccb07c41a5a8f/cobertura/src/site/htdocs/xml/coverage-04.dtd#L4) that was considered in [initial PR](https://github.com/aantron/bisect_ppx/issues/326#issuecomment-647696263).

Would you be ok with such contribution? @aantron 

Thanks!

Issue #326, opened 2020-06-22
Support Cobertura output format


arvidj:
The [Gitlab code coverage visualization](https://docs.gitlab.com/ee/user/project/merge_requests/test_coverage_visualization.html) uses the [Cobertura](https://cobertura.github.io/cobertura/) format. Adding this type of output would ease simplify coverage visualization for OCaml projects.

aantron:
For future reference, here is [example Cobertura output](https://gist.github.com/apetro/fcfffb8c4cdab2c1061d) and the [DTD](http://cobertura.sourceforge.net/xml/coverage-04.dtd). It seems considerably more complex that Coveralls format, and seems to make more assumptions about the source language.

aantron:
[This example](https://istanbul.js.org/docs/advanced/alternative-reporters/#cobertura) might help with adapting Cobertura to OCaml.

aantron:
Closing this issue as I won't have time to do it myself. @arvidj if you (or anyone else) would like to open a PR about this, I will be happy to review and merge it.

sir4ur0n:
Assuming someone would be interested to give it a shot, do you have any advice/pointers/hints on how to proceed @aantron please? Ideally for someone who has not yet developed in `bisect_ppx` :grin:

aantron:
It looks like you should be able to start by duplicating the [Coveralls formatter](https://github.com/aantron/bisect_ppx/blob/a72c5f244aa474f4110821943ce6e54c1f65bff6/src/report/report_coveralls.ml) and then modifying it.

If you work on this, please ping here, as I have plans to potentially work on this in the relatively near term, as well.

sir4ur0n:
If you have plans to work on it I gladly leave the floor to you, as I think you would be significantly more efficient than me, considering you already know the code base :smile:

vch9:
Just wanted to let you know I am starting to work on this feature!

PR #420, opened 2023-06-11
Add <packages> to cobertura format

maxim092001:
This PR adds `<packages>` to cobertura format.

[Issue](https://github.com/aantron/bisect_ppx/issues/416)

### DIFF
--- a/src/report/cobertura.ml
+++ b/src/report/cobertura.ml
@@ -21,23 +21,27 @@ type package = {
   classes : class_ list;
 }
 
+type packages = {
+  packages : package list;
+}
+
 type cobertura = {
   lines_valid : int;
   lines_covered : int;
   line_rate : float;
   sources : string list;
-  package : package;
+  packages : packages;
 }
 
 let pp_list pp fmt =
   List.iter (fun x -> pp fmt x; Format.pp_print_string fmt "\n")
 
 let pp_line fmt {number; hits} =
-  Format.fprintf fmt "          <line number=\"%d\" hits=\"%d\"/>" number hits
+  Format.fprintf fmt "            <line number=\"%d\" hits=\"%d\"/>" number hits
 
 let pp_lines fmt lines =
   let open Format in
-  fprintf fmt "        <lines>\n%a        </lines>\n"
+  fprintf fmt "          <lines>\n%a          </lines>\n"
     (pp_list pp_line) lines
 
 let pp_class_ fmt {name; line_rate; lines} =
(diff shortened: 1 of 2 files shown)


aantron:
Thank you!

@arvidj, @sir4ur0n, @vch9, would this be fine for your usage of the Cobertura reports, which is, I understand, in GitLab CI? I think you could directly try it by using Bisect_ppx from this branch, `https://github.com/maxim092001/bisect_ppx.git#master`.

> Hey!
> 
> ## Proposal:
> I would like to add `<packages>` around packages.
> 
> For example current report generated by bisect_ppx looks like:
> 
> ```xml
> <?xml version="1.0"?>
> <coverage lines-valid="282" lines-covered="79"
> line-rate="0.280142">
>   <sources>
>     <source>.</source>
>   </sources>
>   <package name="." line-rate="0.280142">
>     <classes>
>       <class name="lib/message.ml" filename="lib/message.ml"
>       line-rate="0.250000">
>         <lines>
>           <line number="4" hits="4" />
>           <line number="6" hits="0" />
>           <line number="7" hits="0" />
>           <line number="8" hits="0" />
>         </lines>
>       </class>
>     </classes>
>   </package>
> </coverage>
> ```
> 
> And I would like it to be as:
> 
> ```xml
> <?xml version="1.0"?>
> <coverage lines-valid="282" lines-covered="79"
> line-rate="0.280142">
>   <sources>
>     <source>.</source>
>   </sources>
>   <packages>
>     <package name="." line-rate="0.280142">
>       <classes>
>         <class name="lib/message.ml" filename="lib/message.ml"
>         line-rate="0.250000">
>           <lines>
>             <line number="4" hits="4" />
>             <line number="6" hits="0" />
>             <line number="7" hits="0" />
>             <line number="8" hits="0" />
>           </lines>
>         </class>
>       </classes>
>     </package>
>   </packages>
> </coverage>
> ```
> 
> ## Why:
> Some tools used to generate HTML or any other representation of report require `<packages>`. For example: [pycobertura](https://github.com/aconrad/pycobertura/tree/master)
> 
> Also, in [main cobertura repo](https://github.com/cobertura/cobertura/blob/0ff963284cecaace30f409a977dccb07c41a5a8f/cobertura/src/main/java/net/sourceforge/cobertura/reporting/xml/XMLReport.java#L138) they also have `<packages>`.
> 
> Also, `<packages>` is a part of [some kind of official specification](https://github.com/cobertura/cobertura/blob/0ff963284cecaace30f409a977dccb07c41a5a8f/cobertura/src/site/htdocs/xml/coverage-04.dtd#L4) that was considered in [initial PR](https://github.com/aantron/bisect_ppx/issues/326#issuecomment-647696263).
> 
> Would you be ok with such contribution? @aantron
> 
> Thanks!

aantron:
Thank you! I will release this shortly. I guess we will find out retroactively if this breaks anything, but I expect that it won't, because the `<packages>` element appears to be standard.

maxim092001:
> Thank you! I will release this shortly. I guess we will find out retroactively if this breaks anything, but I expect that it won't, because the `<packages>` element appears to be standard.

Thanks! I will close the issue then.

arvidj:
Hello, a bit late to the party, sorry about that. I've tested the bisect_ppx version from this branch and it seems to work fine in our setup. Thanks for the patch!

PR #421, opened 2023-06-21
Fix stack overflow in HTML generation

AllanBlanchard:
I found a stack overflow when running bisect on the code base of Frama-C. This PR replaces a try catch with a match exception to avoid stacking try contexts. ~~I would probably be better to make the function tail recursive though, but I don't know when I can do it.~~ Ok, I miss-used the `[@tailcall]`, now it *is* tailrec.

### DIFF
--- a/src/report/html.ml
+++ b/src/report/html.ml
@@ -339,8 +339,9 @@ let output_for_source_file
     let lines, line_count =
       let rec read number acc =
         let start_ofs = pos_in in_channel in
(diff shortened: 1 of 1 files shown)


AllanBlanchard:
Ok, the CI  has the same behavior as on the `master` branch so I guess we are OK here.

aantron:
Thank you! This will be included in the next release.

Issue #434, opened 2024-02-10
Question on out-edge instrumentation of if-then-else and match cases


mbarbin:
I am currently integrating `bisect_ppx` into a code base to monitor coverage as I
add more tests. During this process, I've noticed some instances where
`bisect_ppx` reports lines as uncovered, even though I believe they should not
be. I'm hoping to gain a better understanding of the instrumentation strategy
and consider whether a slight modification to the instrumentation heuristic
might be warranted.

The specific scenario I'm focusing on involves the out-edge instrumentation of
if-then-else and match cases, which I understand is handled by the
`___bisect_post_visit___` construct.

In the `bisect_ppx` tests, I found these examples:

`test/instrumentation/control/if.t`:
```ocaml
  let _ =
    if bool_of_string "true" then (
      ___bisect_visit___ 3;
      ___bisect_post_visit___ 0 (print_endline "foo"))
    else (
      ___bisect_visit___ 2;
      ___bisect_post_visit___ 1 (print_endline "bar"))
  
  let _ =
   fun () ->
    ___bisect_visit___ 6;
    if bool_of_string "true" then (
      ___bisect_visit___ 5;
      print_endline "foo")
    else (
      ___bisect_visit___ 4;
      print_endline "bar")
```

`test/instrumentation/control/match.t`:
```ocaml
 let _ =
    match print_endline "foo" with
    | () ->
        ___bisect_visit___ 1;
        ___bisect_post_visit___ 0 (print_endline "bar")
  
  let _ =
   fun () ->
    ___bisect_visit___ 3;
    match print_endline "foo" with
    | () ->
        ___bisect_visit___ 2;
        print_endline "bar"
```

Both examples include a case where the branches are post-instrumented (shown
first) and a case where they are not (shown second).

Could we discuss examples where post-instrumentation is beneficial in these
scenarios? I'm unsure. When the expression in the case is a `Pexp_apply`, and
the function raises an exception, I don't believe this should be considered as
uncovered.

The specific case that led me to investigate this is as follows[^1]:

`src/status_line/ml`:
```ocaml
let create ~size ~code =
  if code < 0 then raise_s [%sexp "invalid negative code", { size : int; code : int }];
  (* Do more stuff below *)
```

I examined the instrumented code using `dune describe`:

```sh
$ dune describe pp my_file.ml --instrument-with bisect_ppx
```

which shows:

```ocaml
let create ~size  ~code  =
  ___bisect_visit___ 31;
  if code < 0
[...]

PR #433, opened 2024-02-10
Add `--sort-by-stats` flag to html reporter

mbarbin:
In this pull request, I have added a `--sort-by-stats` flag to the html-report
command. This flag provides an option for users to order the files in the index
page of the report by their coverage statistics in ascending order.

I have also ensured that this new flag works in conjunction with the existing
`--tree` option : when both `--sort-by-stats` and `--tree` flags are used, the
directories, as well as the files within each directory, are sorted by coverage
statistics.

This addition is intended to offer a different perspective for developers who
are working on improving coverage in existing codebases. By presenting files
with lower coverage at the top, this option could potentially assist in
prioritizing areas that may benefit from additional testing.

### Note on Code Formatting

I encountered some difficulties to `fmt` the code due to issues finding a
compatible combination of dependencies with opam. Despite trying several OCaml
switches ranging from 4.05 to 5.1, I was unsuccessful. The `ocamlformat.0.16.0`
version seems to be outdated. Any assistance in resolving this issue would be
greatly appreciated!

Looking forward to your feedback!

PR #435, opened 2024-02-12
Add tests to document current behavior of post-instr in branches

mbarbin:
This PR adds new test cases to monitor the current behavior of branch
instrumentation in Bisect_ppx. The test cases focus on complex scenarios
involving if-then-else and match branches discussed in #434.

The aim is not to modify the existing behavior directly, but to establish a
baseline for future changes.

By adding these tests, we can monitor how future changes to Bisect_ppx may alter
these behaviors (TDD).

### DIFF
--- a/test/instrument/control/if.t
+++ b/test/instrument/control/if.t
@@ -97,3 +97,116 @@ tail position iff the whole if-expression is in tail position.
     else (
       ___bisect_visit___ 4;
       print_endline "bar")
+
+When the if-then-else is in a sequence, the next-expressions are
+instrumented as expected.
+
+  $ bash ../test.sh <<'EOF'
+  > let _ =
+  >   if bool_of_string "true" then print_endline "foo" else print_endline "bar";
+  >   print_endline "this";
+  >   print_endline "that";
+  > EOF
+  let _ =
+    if bool_of_string "true" then (
+      ___bisect_visit___ 5;
+      ___bisect_post_visit___ 2 (print_endline "foo"))
+    else (
+      ___bisect_visit___ 4;
+      ___bisect_post_visit___ 3 (print_endline "bar"));
+    ___bisect_post_visit___ 1 (print_endline "this");
+    ___bisect_post_visit___ 0 (print_endline "that")
+
+Where there's a raise or a failwith guarded by a conditional, the
+branch itself is instrumented, but the raising expression is not
+post-instrumented...
+
+  $ bash ../test.sh <<'EOF'
+  > let _f ~cond1 ~cond2 =
+  >   if cond1 then failwith "this";
+  >   if cond2 then raise Not_found;
+  >   print_endline "this";
+  >   print_endline "that";
+  > EOF
+  let _f ~cond1 ~cond2 =
+    ___bisect_visit___ 5;
+    if cond1 then (
+      ___bisect_visit___ 4;
+      failwith "this");
+    ___bisect_visit___ 3;
+    if cond2 then (
+      ___bisect_visit___ 2;
+      raise Not_found);
+    ___bisect_visit___ 1;
+    ___bisect_post_visit___ 0 (print_endline "this");
+    print_endline "that"
+
+... however note that this doesn't hold for all raising expressions.
+For example, if you use `invalid_arg`, or a raising helper, its
+application will be post-instrumented (unless all of its arguments are
+labeled !?), resulting in a code location that can never be visited,
+by design. These are currently acknowledged limitations.
+
+  $ bash ../test.sh <<'EOF'
+  > let invalid_n_labeled ~n = invalid_arg (Printf.sprintf "Invalid value of n=%d" n)
+  > 
+  > let invalid_p_unlabeled p = invalid_arg (Printf.sprintf "Invalid value of p=%d" p)
+  > 
+  > let _f ~cond1 ~n ~p =
+  >   if cond1 then invalid_arg "this";
+  >   if n < 0 then invalid_n_labeled ~n;
+  >   if p < 0 then invalid_p_unlabeled p;
+  >   print_endline "this";
+  >   print_endline "that";
+  > EOF
+  let invalid_n_labeled ~n =
+    ___bisect_visit___ 1;
+    invalid_arg
+      (___bisect_post_visit___ 0 (Printf.sprintf "Invalid value of n=%d" n))
+  
+  let invalid_p_unlabeled p =
+    ___bisect_visit___ 3;
+    invalid_arg
+      (___bisect_post_visit___ 2 (Printf.sprintf "Invalid value of p=%d" p))
+  
+  let _f ~cond1 ~n ~p =
+    ___bisect_visit___ 13;
+    if cond1 then (
+      ___bisect_visit___ 12;
+      ___bisect_post_visit___ 11 (invalid_arg "this"));
+    ___bisect_visit___ 10;
+    if n < 0 then (
+      ___bisect_visit___ 9;
+      invalid_n_labeled ~n);
+    ___bisect_visit___ 8;
+    if p < 0 then (
+      ___bisect_visit___ 7;
+      ___bisect_post_visit___ 6 (invalid_p_unlabeled p));
+    ___bisect_visit___ 5;
+    ___bisect_post_visit___ 4 (print_endline "this");
+    print_endline "that"
+
+If you try and disable post-instrumentation with a `coverage off`
+directive, the branch pre-instrumentation is removed as well,
+resulting in losing the ability to control that the branch itself is
+visited.
+
+  $ bash ../test.sh <<'EOF'
+  > let invalid_p_unlabeled p = invalid_arg (Printf.sprintf "Invalid value of p=%d" p)
+  > 
+  > let _f ~p =
+  >   if p < 0 then (invalid_p_unlabeled p [@coverage off]);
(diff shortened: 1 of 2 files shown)


test/instrument/control/if.t
```diff
@@ -97,3 +97,116 @@ tail position iff the whole if-expression is in tail position.
...
+      ___bisect_post_visit___ 2 (print_endline "foo"))
+    else (
+      ___bisect_visit___ 4;
+      ___bisect_post_visit___ 3 (print_endline "bar"));
+    ___bisect_post_visit___ 1 (print_endline "this");
+    ___bisect_post_visit___ 0 (print_endline "that")
+
+Where there's a raise or a failwith guarded by a conditional, the
+branch itself is instrumented, but the raising expression is not
+post-instrumented...
```
aantron, line 122: What is the incremental value of having this test over the tests in `apply/special.t`, including https://github.com/aantron/bisect_ppx/blob/e30265643e77bcf2c9eba7322429c779122106fc/test/instrument/apply/special.t#L293 ? Is there any need to test this *in interaction with* `if`-expressions?

test/instrument/control/if.t
```diff
@@ -97,3 +97,116 @@ tail position iff the whole if-expression is in tail position.
...
+      raise Not_found);
+    ___bisect_visit___ 1;
+    ___bisect_post_visit___ 0 (print_endline "this");
+    print_endline "that"
+
+... however note that this doesn't hold for all raising expressions.
+For example, if you use `invalid_arg`, or a raising helper, its
+application will be post-instrumented (unless all of its arguments are
+labeled !?), resulting in a code location that can never be visited,
+by design. These are currently acknowledged limitations.
```
aantron, line 148: This is an acknowledged and inevitable limitation because there is no way for Bisect to know that a specific identifier refers to a function that always or never raises. In fact, it doesn't know that for the identifiers `failwith` or `invalid_arg` either, as those can be redefined and shadowed by the user, including by another module or a third-party library that is not even processed by Bisect during compilation of the user's particular project. I don't see that we need a test for this -- the actual behavior is covered by the normal `apply` tests.

test/instrument/control/if.t
```diff
@@ -97,3 +97,116 @@ tail position iff the whole if-expression is in tail position.
...
+      ___bisect_visit___ 7;
+      ___bisect_post_visit___ 6 (invalid_p_unlabeled p));
+    ___bisect_visit___ 5;
+    ___bisect_post_visit___ 4 (print_endline "this");
+    print_endline "that"
+
+If you try and disable post-instrumentation with a `coverage off`
+directive, the branch pre-instrumentation is removed as well,
+resulting in losing the ability to control that the branch itself is
+visited.
```
aantron, line 192: Could you move this to [`attribute.t`](https://github.com/aantron/bisect_ppx/blob/e30265643e77bcf2c9eba7322429c779122106fc/test/instrument/attribute.t)? That is the place for testing `[@coverage off]`. It already includes cases for interaction of `[@coverage off]` with multiple language constructs in case.

test/instrument/control/match.t
```diff
@@ -65,3 +65,136 @@ cases are in tail position iff the match expression is in tail position.
     | () ->
         ___bisect_visit___ 2;
         print_endline "bar"
+
```
aantron, line 68: Analogous questions for the `match.t` changes as for the `if.t` changes.

test/instrument/control/if.t
aantron: (Raises or diverges)

test/instrument/control/if.t
mbarbin: I don't believe there's a need for this test from the standpoint of the current
implementation. I do see its value in terms of future-proofing and documenting.

The `if cond then raise` pattern is quite common. As it stands, one must be
cautious with how the raising expression is written. It's important to note that
even seemingly innocuous refactoring, which shouldn't affect coverage, could
potentially introduce non-visitable coverage points. This can be quite
surprising.

Consider the following example:

```diff
-| if cond then raise Not_found
+| if cond then raise_s [%sexp "Key not found", { key : Key.t}]
```

This results in the insertion of non-visitable point 1:

```diff
-| if cond then raise Not_found
+| if cond then (__bisect_post__visit 1 (raise_s [%sexp "Key not found", { key : Key.t}]))
``

test/instrument/control/if.t
mbarbin: > there is no way for Bisect to know that a specific identifier refers to a
> function that always or never raises

I agree. It's fair to say that "a strategy to limit non-visitable points, which
relies partly on identifiers, will inevitably encounter these limitations."
However, if we were to discover an alternative approach (assuming one exists),
we might still see improvements in this test. In other words, the insertion of
post-instrumentation might not be inevitable when considering the range of
possible instrumentation strategies.

mbarbin: > (Raises or diverges)

Btw, I didn't understand the part where you said  "or diverges".

aantron:
> Perhaps you find the cost of adding such tests outweighs the potential benefits,
> considering the risk that such improvements might not be feasible? If you prefer
> we can wait to see more progress on an overall proposal before deciding on
> whether to pursue merging these?

I would suggest to keep the tests that are redundant right now locally, and to include them as part of the PR if you do suggest an alternative instrumentation strategy for `if`+`Pexp_apply`. Normally when I review such PRs, I manually run the new tests with the base branch to observe the difference with the PR, so I will be able to see the difference. That is, if I don't spot it by inspection, which is quite likely in this case, I think.

test/instrument/control/if.t
mbarbin: I appreciate your suggestion to add tests to `attribute.t`, which I had
overlooked. I will proceed to incorporate new tests there to demonstrate how to
disable coverage in branches (same PR ok?).

Despite the presence of branch coverage testing in `attribute.t`, I advocate for
retaining this specific part of the control test here. It serves to illustrate
that this method does not provide a viable workaround for the issue at hand.
This was in response to your comment in #433:

> I believe in your case you should suppress instrumentation of the out-edge
> with [@coverage off], though I don't immediately recall whether the way that
> is implemented will also cause the branch instrumentation to also be
> supressed.

This might come across more as literate testing (akin to a tutorial, perhaps?)
rather than a pure behavior test. I've attempted to construct a small narrative
around the added tests. However, if you have strong reservations, I'm open to
removing this section entirely.

test/instrument/control/if.t
aantron: > > (Raises or diverges)
> 
> Btw, I didn't understand the part where you said "or diverges".

Post-instrumentation checks that a function evaluates to a value. The two ways that it doesn't in OCaml are when it raises an exception or when it diverges.

aantron: Those are the two ways that the out-edge of a`Pexp_apply` may *not* be followed.

aantron:
> Consider the following example:
> 
> ```diff
> -| if cond then raise Not_found
> +| if cond then raise_s [%sexp "Key not found", { key : Key.t}]
> ```
> 
> This results in the insertion of non-visitable point 1:
> 
> ```diff
> -| if cond then raise Not_found
> +| if cond then (__bisect_post__visit 1 (raise_s [%sexp "Key not found", { key : Key.t}]))
> ``
> ```

It may be easier to address this by extending the [`--exclusions` option](https://github.com/aantron/bisect_ppx/blob/e30265643e77bcf2c9eba7322429c779122106fc/doc/advanced.md#excluding-generated-files-from-coverage) to allow the user to define a list of identifiers to exclude from coverage as if they were built-ins when they appear in `Pexp_apply`.

mbarbin:
> I would suggest to keep the tests that are redundant right now locally, and to include them as part of the PR if you do suggest an alternative instrumentation strategy for `if`+`Pexp_apply`. Normally when I review such PRs, I manually run the new tests with the base branch to observe the difference with the PR, so I will be able to see the difference. That is, if I don't spot it by inspection, which is quite likely in this case, I think.

Understood. I am more accustomed to first publishing tests and then reviewing
changes in subsequent updates. However, I'm more than willing to adapt as
necessary. I'll transition the PR to the draft stage as I dedicate more time to
the strategy. Thank you!

aantron:
> Understood. I am more accustomed to first publishing tests and then reviewing
> changes in subsequent updates. However, I'm more than willing to adapt as
> necessary. I'll transition the PR to the draft stage as I dedicate more time to
> the strategy. Thank you!

Another way to do this is to open the PR with the tests in the first commit and the other changes afterwards, with the changes in test output included in the later commits. I suppose you are potentially turning this PR into that right now :) Thank you.

test/instrument/control/if.t
mbarbin: Does "diverging" encompass non-termination and potentially the effects in OCaml 5? Something else?

Regarding the effects, I'm optimistic that their influence on the strategy will
be similar to that of exceptions. I hope this won't necessitate too big of an
exploration of the OCaml 5 runtime :exploding_head:. Thank you again for your explanation!

aantron: Yes, "diverging" is non-termination. OCaml 5 effects are supposed to always be returned from exactly once, so they should not affect control flow as visible to Bisect, although this is not enforced, and for effects that correspond to async computation the "exactly once" might be later than program termination, so in that sense they behave like exceptions.

PR #436, opened 2024-03-20
Add support for automatic uploading to gitlab

JasonGross:


src/report/coveralls.ml
```diff
@@ -220,12 +225,14 @@ struct
     match ci, service with
     | `CircleCI, `Coveralls -> Some "CIRCLE_PULL_REQUEST"
     | `GitHub, `Coveralls -> Some "PULL_REQUEST_NUMBER"
+    | `GitLab, `Coveralls -> Some "CI_EXTERNAL_PULL_REQUEST_IID"
```
JasonGross, line 228: Not quite sure if this is right, but it's what's listed [here](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html)

